PR #2, opened 2017-03-12
Do not mark attribute as changed if value is the same

shioyama:
(Note: I'm starting to post bug fixes and important changes as pull requests to
make them more visible.)

Currently, for a model post with dirty tracking enabled, this is what happens:

```ruby
post = Post.new
post.title        #=> nil
post.changed?     #=> false
post.changed      #=> []
post.changes      #=> {}
```

So far so good, but if we set the value to `nil`:

```ruby
post.title = nil
post.changed?     #=> true
```

This is a bug, which this PR fixes. After the change:

```ruby
post.title = nil
post.changed?   #=> false
post.title = "foo"
post.changed?   #=> true
```

Note: the change applies to both ActiveModel and Sequel modules.

### DIFF
--- a/lib/mobility/backend/active_model/dirty.rb
+++ b/lib/mobility/backend/active_model/dirty.rb
@@ -27,7 +27,7 @@ value of the translated attribute if passed to it.
         locale_accessor = "#{attribute}_#{locale}"
         if model.changed_attributes.has_key?(locale_accessor) && model.changed_attributes[locale_accessor] == value
           model.attributes_changed_by_setter.except!(locale_accessor)
-        elsif read(locale, **options) != value
+        elsif read(locale, options.merge(fallbacks: false)) != value
           model.send(:attribute_will_change!, "#{attribute}_#{locale}")
         end
         super
--- a/lib/mobility/backend/active_model/dirty.rb
+++ b/lib/mobility/backend/active_model/dirty.rb
@@ -23,11 +23,11 @@ value of the translated attribute if passed to it.
     module ActiveModel::Dirty
       # @!group Backend Accessors
       # @!macro backend_writer
-      def write(locale, value, **)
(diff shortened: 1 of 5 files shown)


shioyama:
While working on this fix I noticed that there is a tricky relationship between dirty tracking and fallbacks. If I have a model `Article` with fallbacks enabled:

```ruby
class Article
  include Mobility
  translates :title, dirty: true, fallbacks: { en: 'ja' }
end
```

and I set the value in Japanese, the value in English will fall through to this value:

```ruby
article = Article.new
article.title
#=> nil
Mobility.with_locale(:ja) { article.title = "タイトル" }
article.title
#=> "タイトル"
```

In the dirty module, when we check the current value of the attribute in the current locale, we need to disable fallbacks, otherwise we will compare with the fallback value and mark a change when we set to the same blank value:

```ruby
article.title = nil
article.changed
#=> ["title_ja", "title_en"]
```

We don't want this - i.e., while I may want to fall through to a different locale, I don't want to compare with the value in that locale when marking attributes in the current locale as having changed or not.

If instead we disable fallbacks (`read(locale, options.merge(fallbacks: false))`), then we get what (I think) you'd expect:

```ruby
article.title = nil
article.changed
#=> ["title_ja"]
```

Only the value in Japanese has changed (from `nil` to `タイトル`).

However, when we *do* update the value, do we mark the change as a change from `nil` or as a change from the fallback value?

i.e.

```ruby
article.title = "title"
article.changes
#=> { "title_ja" => [nil, "タイトル"], "title_en" => ["タイトル", "title"] }
```

Or should it be marked as a change from `nil` to the new value?

I have gone with the former (change from fallback value), becuase: a) to make it do otherwise would require hacking hte internals of Rails/Sequel, which I don't want to do, b) I think intuitively, this sort of makes sense.

Issue #5, opened 2017-03-15
Pass fallback locale in getter


simonknoll:
Accessing a specific language works already neatly with
`post.title(locale: :en)`
It would be great if one could simply pass a fallback locale in the getter directly
`post.title(locale: :en, fallback: :de)` #

shioyama:
Interesting idea! I'll have a look.

simonknoll:
:+1: thanks for considering!

shioyama:
@simonknoll see #9.

simonknoll:
Thank you so much, this makes my live much easier!

shioyama:
#9 has been merged, closing this.

simonknoll:
@shioyama I tried now with version `0.1.7` but I can't get it working. Maybe I'm using it not as intended. Anyways I've created a simple test project with a [rspec test](https://github.com/simonknoll/mobility-fallback/blob/master/spec/models/contribution_spec.rb). This is the repository https://github.com/simonknoll/mobility-fallback

Cheers,
Simon

shioyama:
@simonknoll I see the problem. You need to enable fallbacks on the model in order to set them in the getter, otherwise the `Mobility::Backend::Fallbacks` module will not be included.

So if you define the `Contribution` model this way, it will work:

```ruby
class Contribution < ApplicationRecord
  include Mobility
  translates :landing_page, type: :string, fallbacks: true
end
```

Note that if you then want somewhere else *not* to use fallbacks for `landing_page`, you'll need to pass `fallback: false`, i.e. `contribution.landing_page(fallback: false)`.

We probably need a way to include the module, but disable it by default. I'm thinking about a natural interface for that...

shioyama:
Actually I'm noticing an issue with 0.1.7 which is causing lots of bogus warnings in the logs. I'll fix it tomorrow and release a new version. (Working on this on [this](https://github.com/shioyama/mobility/tree/fallbacks_super) branch.)

Also, I should be able to change the API so that it includes the fallbacks module by default, but only actually uses it if when `fallback: ...` is passed to the getter. But I want to think about this a bit more before making that change.

simonknoll:
Thank you for the clarification @shioyama!

shioyama:
@simonknoll try version 0.1.10 (had to yank 0.1.8 + 0.1.9...).

PR #9, opened 2017-03-16
Allow passing fallback locale or locales to getter method

shioyama:
See #5.

This is a slight API change, previously `post.title(fallbacks: false)` would disable fallbacks for the fetch, but thinking about this I find `post.title(fallback: false)` (singular `fallback`) makes more sense since the option is specifying *not to fall back*.

In addition, with this PR, the option now accepts one or more locales its value, and when present will use these instead of the default fallbacks or fallbacks defined for the attribute in the model.

So:

```ruby
post.title(fallback: :fr)
```

will fall back to the French locale if the value in the current locale is nil.

```ruby
post.title(fallback: [:fr, :es])
```

will fall back first to French, then (if that is empty) to Spanish, if the value in the current locale is nil.

This can be combined with the `locale` option to specify both locale to fetch and fallbacks, e.g.:

```ruby
post.title(locale: :de, fallback: [:fr, :es])
```

would try first German, then French, then Spanish, *regardless of what locale we are currently in or what fallbacks are defined on the attribute at the model level*.

### DIFF
--- a/lib/mobility/backend.rb
+++ b/lib/mobility/backend.rb
@@ -76,12 +76,11 @@ On top of this, a backend will normally:
     # @!macro [new] backend_constructor
     #   @param model Model on which backend is defined
     #   @param [String] attribute Backend attribute
-    #   @option options [Hash] fallbacks Fallbacks hash
-    def initialize(model, attribute, **options)
+    #   @param [Hash] fallbacks Fallbacks hash
+    def initialize(model, attribute, fallbacks: nil, **options)
       @model = model
       @attribute = attribute
       @options = options
-      fallbacks = options[:fallbacks]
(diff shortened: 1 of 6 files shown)


simonknoll:
Hi @shioyama, unfortunately i can't get it running
in my testcase I have only the :de translation, so 
`contribution.landing_page(locale: :de)` returns correctly  `"mypage.de"`
unfortunately
`contribution.landing_page(locale: :fr, fallback: [:de, :en])` returns `nil`

Am i using the fallback functionality in the wrong way?

shioyama:
Are you using the version on master? I haven't released it yet.

shioyama:
Ok I released 0.1.7, please try updating to that.

simonknoll:
I was using once the master branch and once the feature branch, both with the same result. Will try now the release 0.1.7. Thanks!

PR #15, opened 2017-04-02
Destroy all translations after model is destroyed (KeyValue backend)

shioyama:
This is a small but important change to ensure data integrity. When you translate a model using the KeyValue backend with some attribute, say `title`, then later decide to change the name of the attribute to something else, your earlier translations will disappear. This is natural since they were stored with the key `title`, and Mobility is now looking for a different key on the association.

This means that if you change your attribute name *back*, those earlier translations will again become available, because they were never destroyed (the change of attribute name was made at the code level, not the data level). I consider this a good thing.

However, currently they will also remain in the db even after you *destroy* the model, although translations with keys currently defined on the model will be destroyed. The reason is that there is a `dependent: :destroy` option on the translations association, but the association itself is limited with a `where(key: attributes)` scope. The scope is important for efficiency, to avoid looping through "dead" translations that are no longer used, but it also results in these translations not being seen as dependencies.

To fix this, I've added after destroy callbacks for ActiveRecord and Sequel which destroys *all* translations associated with this model, regardless of their key value. This means that *provided you have at least one KeyValue attribute defined on the model when you destroy it*, all translations will be destroyed as you would expect. I think this is reasonable and the best we can do.

### DIFF
--- a/lib/mobility/backend/active_record/key_value.rb
+++ b/lib/mobility/backend/active_record/key_value.rb
@@ -1,3 +1,5 @@
+# frozen-string-literal: true
+
 module Mobility
   module Backend
 =begin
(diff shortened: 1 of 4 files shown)


garshyn:
I have a model with `acts_as_paranoid` ([`paranoia`](https://github.com/rubysherpas/paranoia) gem) and Mobility modules included with `KeyValue` backend.

When a record is soft deleted, it remains in the database, but the translations are gone.

Is it possible to keep the translations?

shioyama:
Hmm... well, this could easily be made an option for the KeyValue backend. Say:

```ruby
  translates :foo, dependent_destroy: false
end
```

Or something like this. Can you make a feature request (issue) for it?

PR #18, opened 2017-04-04
Drop its

sachin21:
Dropped its. If mobility once uses, I think `its` is needless.


shioyama:
Thanks! Will merge once specs pass.

sachin21:
@shioyama Thanks for quickly merging! 😄

Issue #33, opened 2017-05-24
Uniqueness validator returns exists for non existing records


jonian:
Hi, thank you for this great gem.

When I use `uniqueness: true` in validations, I get the following message. I have tried using it with an existing database and a new database.

    Running via Spring preloader in process 16712
    Loading development environment (Rails 5.1.1)
    [1] pry(main)> Category.create(name: 'CSR')
       (0.2ms)  BEGIN
      Mobility::ActiveRecord::StringTranslation Exists (1.3ms)  SELECT  1 AS one FROM "mobility_string_translations" WHERE "mobility_string_translations"."key" = $1 AND "mobility_string_translations"."translatable_id" IS NULL AND "mobility_string_translations"."translatable_type" = $2 AND "mobility_string_translations"."locale" = $3 LIMIT $4  [["key", "name"], ["translatable_type", "Category"], ["locale", "en"], ["LIMIT", 1]]
       (0.2ms)  ROLLBACK
    ActiveRecord::ImmutableRelation: ActiveRecord::ImmutableRelation
    from /home/jonian/.gem/ruby/2.4.0/gems/activerecord-5.1.1/lib/active_record/relation/query_methods.rb:936:in `assert_mutability!'

I am using the `Key-Store` backend, `PostgreSQL 9.6.3`, `Rails 5.1.1` and `Ruby 2.4.1`.
The model configuration is below:

    class Category < ApplicationRecord
      # Include modules
      include Mobility

      # Translate attributes
      translates :name, type: :string, locale_accessors: true, fallbacks: true

      # Attribute validations
      validates :name, presence: true, uniqueness: true
    end

    class ApplicationRecord < ActiveRecord::Base
      # Set class as abstract
      self.abstract_class = true
    end

jonian:
It seems that the problem is in `mobility/lib/mobility/active_record/uniqueness_validator.rb` line `14`.

    if relation.exists?
      error_options = options.except(:case_sensitive, :scope, :conditions)
      error_options[:value] = value

      record.errors.add(attribute, :taken, error_options)
    end

Using `relation.exists?` or `relation.any?` I get the same error. If i use `relation.present?`, it works as expected.

shioyama:
Ok I found the root cause.

[...]

PR #35, opened 2017-05-24
Fix uniqueness validation error

jonian:
This fixes issue #33

### DIFF
--- a/lib/mobility/active_record/uniqueness_validator.rb
+++ b/lib/mobility/active_record/uniqueness_validator.rb
@@ -11,7 +11,7 @@ module Mobility
           relation = mobility_scope_relation(record, relation)
           relation = relation.merge(options[:conditions]) if options[:conditions]
 
-          if relation.exists?
+          if relation.present?
             error_options = options.except(:case_sensitive, :scope, :conditions)
             error_options[:value] = value
 


shioyama:
Thanks very much! I'm going to enable specs for Rails 5.1+ and see what else is failing. I want to see if this is an issue with query methods in general rather than the uniqueness validator in particular -- if that's the case, I'd prefer to fix the root cause.

shioyama:
I've started running tests against Rails 5.1 in the [active_record_5.x](https://github.com/shioyama/mobility/tree/active_record_5.x) branch, and as expected there are a few issues, many deprecation warnings and some failing specs. I'd like to handle these together, let me have a look and get back to this.

If you want to have a look for yourself, use the branch and `export RAILS_VERSION=5.1` before running specs. There are about 40 failing, I think all with the same thing.

shioyama:
So, as I suspected, this is a problem in Rails 5.1 with `exists?` on queries with translated attributes. There are no actual specs for this method right now but adding one will make this clear.

I'd prefer to fix the root cause (fix `exists?`), so I'm going to investigate that. `present?` converts the query into an array and checks if it is empty, which gets around the problem, so as a quick fix please use it for your project but I'll try to get a proper fix for 5.1 out soon.

shioyama:
Closing this since I fixed the root cause in #36. Thanks again for the PR and the issue!

jonian:
Thanks again for the quick fix!

PR #36, opened 2017-05-27
Avoid using respond_to? on relation in query methods

### LINKED ISSUE #33: Uniqueness validator returns exists for non existing records

shioyama:
Fixes #33 (AR 5.1)

@jonian can you try this branch and confirm it fixes the issue for you?

### DIFF
--- a/lib/mobility/backend/active_record/query_methods.rb
+++ b/lib/mobility/backend/active_record/query_methods.rb
@@ -20,7 +20,7 @@ models. For details see backend-specific subclasses.
         # @note Only want to define this once, even if multiple QueryMethods
         #   modules are included, so define it here in extended method
         def extended(relation)
-          unless relation.respond_to?(:mobility_where_chain)
+          unless relation.methods(false).include?(:mobility_where_chain)
             relation.define_singleton_method(:mobility_where_chain) do
               @mobility_where_chain ||= Class.new(::ActiveRecord::QueryMethods::WhereChain)
             end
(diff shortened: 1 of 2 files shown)


jonian:
@shioyama I tested it and it works without problems. Thank you!

PR #38, opened 2017-05-30
Make generators work with Rails 4.2

lime:
I looked at which tests where failing on the `active_record_4.2` branch, and thought I'd try to help out where I could.

This should fix the failing generator tests on Rails 4.2.

- [x] Use the older `ActiveRecord::Migration` as base class for migrations
- [x] Don't call `primary_key_type` unless it's defined
  - This [seems to be](https://github.com/rails/rails/blob/v5.1.1/activerecord/lib/rails/generators/active_record/migration.rb#L19-L22) used to add `id: :uuid` and similar
  - AFAIK there's no Rails 4.2 equivalent, so we can just leave it out
- [x] Don't call `token?` unless it's defined
  - I'm guessing this is for the new `has_secure_token` attributes
  - Again, it shouldn't need any 4.2 counterpart

Let me know what you think. :)

### DIFF
--- a/lib/generators/rails/mobility/active_record_migration_compatibility.rb
+++ b/lib/generators/rails/mobility/active_record_migration_compatibility.rb
@@ -1,0 +1,14 @@
+require "rails/generators/active_record"
+require "active_record/migration"
+
+module Mobility
+  module ActiveRecordMigrationCompatibility
+    def activerecord_migration_class
+      if ::ActiveRecord::Migration.respond_to?(:current_version)
+        "ActiveRecord::Migration[#{::ActiveRecord::Migration.current_version}]"
+      else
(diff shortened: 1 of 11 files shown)


shioyama:
Wow, thanks so much! This looks great. I'll have a look asap and get back to you.

shioyama:
I just rebased `active_record_4.2` against master, since there were some changes related to migrations in specs for AR 5.1. Can you rebase this PR?

lime:
Rebased!

shioyama:
No worries! I'll restart it later. Anyway some specs will still be failing but that's not a problem (as long as there are *less* failing :smile: ).

shioyama:
Merged, thanks very much!

Issue #42, opened 2017-06-15
Invalid single-table inheritance type


ArturLyapin:
Hello. Thanks for this gem. 
I have one problem, I can't use single-table inheritance with mobility gem.
Isolated reproduction here: https://gist.github.com/ArturLyapin/62669b79a0ab3a57ad2f563644f8e9e4

shioyama:
Ok found the source of this problem, fix in #43. I also discovered a separate issue with the key-value backend with STI that I'll need to fix separately.

shioyama:
Fixed in #43.

ArturLyapin:
Yes, it works.
Thanks for the quick feedback.

PR #43, opened 2017-06-15
Add proper STI support

### LINKED ISSUE #41: ActiveRecord querying not working in > 0.1.14
Behaviour in `0.1.14`
```
[4] pry(main)> User.first
  User Load (1.0ms)  SELECT  "users".* FROM "users" ORDER BY "users"."created_at" ASC LIMIT $1  [["LIMIT", 1]]
=> #<User id: "8ec92fd5-5926-4657-9b32-8cf8382de795", email: "some@example.com", created_at: "2017-05-16 10:19:37", updated_at: "2017-05-16 10:19:37", first_name: nil, last_name: nil, guest: false>
```

Behaviour in `0.1.15` and `0.1.16`
```
[4] pry(main)> User.first
NoMethodError: undefined method `[]' for nil:NilClass
from /Users/<user>/.rvm/gems/ruby-2.4.1/gems/activerecord-5.1.0/lib/active_record/relation/delegation.rb:5:in `relation_delegate_class'
```
Coming from this line: https://github.com/rails/rails/blob/v5.1.0/activerecord/lib/active_record/relation/delegation.rb#L5

I wanted to upgrade the gem, because I was getting a lot of these deprecation warnings:
```
DEPRECATION WARNING: The behavior of `changed_attributes` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_changes.transform_values(&:first)` instead. (called from create_guest_user at /Users/<user>/Desktop/<project>/app/models/user.rb:28)
```
line causing them is just `user.save!`, and a colleague thought that they might be coming from mobility. Seems like this is related https://github.com/rails/rails/pull/28640

shioyama:
Fixes #41 and #42.

### DIFF
--- a/lib/mobility.rb
+++ b/lib/mobility.rb
@@ -248,7 +248,7 @@ module Mobility
     end
 
     def inherited(subclass)
-      subclass.instance_variable_set(:@mobility, @mobility)
+      subclass.instance_variable_set(:@mobility, @mobility.dup)
       super
     end
   end
--- a/lib/mobility.rb
+++ b/lib/mobility.rb
@@ -249,6 +249,7 @@ module Mobility
 
     def inherited(subclass)
       subclass.instance_variable_set(:@mobility, @mobility)
+      super
     end
   end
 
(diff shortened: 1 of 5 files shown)


shioyama:
@goalaleo @ArturLyapin This should fix the STI issues you experienced, and also a couple others that had not yet been reported. Once specs pass I'll merge this and release a new version.

Thanks for the reports!

goalaleo:
@shioyama seems to have worked for me - cheers!

PR #44, opened 2017-06-21
Fix deprecation warnings in Rails 5.1

shioyama:
There are tons of deprecation warnings currently in Rails/AR 5.1. This removes the warnings while still passing specs. However, it introduces some questionable changes to achieve this.

There are two problems:
1. AR::Dirty now requires you to declare all attributes upfront rather than in an open-ended way with `attr_will_change!` (see rails/rails#27963). To get around this, I've overridden `has_attribute?` to make locale accessors (`title_en` etc) act like real attributes. We can't actually define them all since the list of locales is open-ended.
2. You can't call `changes` from within after save callbacks. This conflicts with code used to support `previous_changes` in the Mobility Dirty backend module. This is fixed using a `before_save` hook rather than a `changes_applied` override, but this may fail to work as expected if the model calls `previous_changes` in another `before_save` callback added after this.

The first solution is ok I think for now, the second one may be more problematic but at least is limited to `previous_changes` only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,12 @@ rvm:
   - 2.2.5
   - 2.3.1
 env:
-  - DB=postgres ORM=active_record
-  - DB=mysql ORM=active_record
-  - DB=sqlite3 ORM=active_record
+  - DB=postgres ORM=active_record RAILS_VERSION=5.0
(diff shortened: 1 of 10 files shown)


shioyama:
For the record, I got around the second issue by overriding `changes_internally_applied` in AR >= 5.1.

PR #50, opened 2017-07-08
Reorganize module options

shioyama:
Currently there is a lot of hard-coded dependencies between options-setting in
the `Mobility::Attributes` class and the various backend modules (cache,
fallbacks, etc.), whereas these should really be independent.

I'm going to try to restructure things so that modules can be defined in such a
way that `Mobility::Attributes` does not need to explicitly know anything about
them when including them from options.

### DIFF
--- a/lib/mobility.rb
+++ b/lib/mobility.rb
@@ -170,9 +170,12 @@ module Mobility
     # (see Mobility::Configuration#default_options)
     # @!method default_options
     #
+    # (see Mobility::Configuration#option_modules)
+    # @!method option_modules
+    #
     # (see Mobility::Configuration#default_accessor_locales)
     # @!method default_accessor_locales
-    %w[accessor_method query_method default_backend default_options default_accessor_locales].each do |method_name|
+    %w[accessor_method query_method default_backend default_options option_modules default_accessor_locales].each do |method_name|
       define_method method_name do
         config.public_send(method_name)
       end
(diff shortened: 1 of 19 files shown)


shioyama:
This change adds extracts hard-coded dependencies in the `Mobility::Attributes` class to the different "option modules" (cache, fallbacks, etc) so that the order and set of modules to include can be entirely customized.

Two new configuration options can be used to do this:

```ruby
Mobility.option_modules
```

This is a hash of option keys and constants (classes or modules). The only requirement of the classes used as hash values is that they have a singleton method `apply` which takes an instance if `Attributes` and an option value (typically a boolean) and then does something with the `Attributes` instance, typically including a module into `attributes.backend_class` or into `attributes` itself.

The other configuration option is `Mobility.default_options`. This gets [merged into](https://github.com/lime/mobility/blob/85e4bca4488df89257b62da2e1a204dc795ab99d/lib/mobility/attributes.rb#L121) the options passed into `translates`, thus allowing you to set a hash of defaults to be used across all your models.

valscion:
Thank you for creating these detailed pull requests, and elaborating your thoughts here! Even though it might feel like no-one is reading these, I for one am grateful to be able to follow along how this gem evolves, and how you think through these changes.

🙇

shioyama:
Thanks! That's great to hear. I was mainly doing these PRs as mental notes to myself, with the idea that anybody who was interested could follow along, but very glad to hear that they are useful to you and others. Really appreciate that you took the time to express that here :smile:

PR #52, opened 2017-07-17
Improve how modules are applied

shioyama:
Some more improvements in how modules are applied to a model class and to backend(s).


shioyama:
This is a fairly "deep" refactoring, but the highlights are:

- Use the fact that `Mobility::Backend::Dirty.apply` gets a reference to the `Attributes` instance, which has the model class on it, to avoid some ugly `class_eval` stuff. Basically we can now include any new methods we want on the model by directly including them into the model (from `apply`), in addition to including `read` and `write` overrides to the backend class. (Previously, I used a hook added to the backend's `setup_model` method which was ugly.
- In `Mobility::Attributes`, instead of defining setters/getters and applying option modules in the `initialize` method, define them in `included` instead. By doing this, we know what the model class is, which is important since backends need to know it to know what to include into the backend and model class. Previously the model class was passed in from `translates` but using `includes` is cleaner.
- Moving the remaining module option-specific stuff out of the Attributes specs and into their respective class specs (fallthrough accessors + locale accessors).

PR #68, opened 2017-08-09
Explain write methods behaviour toards persisting translations

madis:
I was writing some acceptance tests when I noticed the translated fields not shown. 

I like the current behaviour where it's up to user to call `save` but the method name kind of implies writing to somewhere. So I think this small adition to docs could avoid some confusion in the future.

If you approve this change, should I also document it elsewhere? I noticed the backends each implement the `write` method so there is no one place to add it. What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,6 +326,10 @@ word.name_backend.read(:en)
 Internally, all methods for accessing translated attributes ultimately end up
 reading and writing from the backend instance this way.
 
+The `write` methods do not call underlying backend's methods to persist the change.
+This is up to the user (e.g. with ActiveRecord you should call `save` write
+the changes to the database.
+
 ### Setting the Locale
 
 It may not always be desirable to use `I18n.locale` to set the locale for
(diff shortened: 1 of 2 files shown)


shioyama:
Thanks for the PR, the change looks good to me! Regarding applying the change to all `write` methods, you could add a line in the `macro` [here](https://github.com/shioyama/mobility/blob/b001a2996240ee217ed75c0fb99ed88caae82ed0/lib/mobility/backend.rb#L71) which is called in all backends, then it will appear in the yard docs on every `write` method.

So you'd add it like this:

```ruby
    # @!macro [new] backend_writer
    #   Some description here.
    #   @param [Symbol] locale Locale to write
    #   @param [Object] value Value to write
    #   @return [Object] Updated value
```

If you want to do that, I'd recommend also adding a description to `read`. Actually looking at these macros now, they should really have descriptions so this would be a great addition.

madis:
Updated. YARD docs now look like

screen shot 2017-08-10 at 12 10 44

shioyama:
Thanks very much!! Merged the changes.

PR #83, opened 2017-09-07
Fix initialization error when inheriting

### LINKED ISSUE #82: Error with initialization
There appears to be a bug on line 257.

https://github.com/shioyama/mobility/blob/07a686309f32da425e3f3e876ec50f16cc3857a1/lib/mobility.rb#L248-L257

Rails was crashing when autoloading a model on `@mobility.dup`, as `@mobility` was nil. Because this is a loading issue, I don't think it is easy to reproduce. I'm guessing you mean `mobility.dup`, which appears to resolve the issue for me.

pwim:
Depending on loading order, `@mobility` may have not been initialized yet, so use method to ensure we aren't trying to dup nil.

Resolves #82.

### DIFF
--- a/lib/mobility.rb
+++ b/lib/mobility.rb
@@ -254,7 +254,7 @@ module Mobility
     end
 
     def inherited(subclass)
-      subclass.instance_variable_set(:@mobility, @mobility.dup)
+      subclass.instance_variable_set(:@mobility, mobility.dup)
       super
     end
   end


shioyama:
Thanks! Will merge once specs pass.

shioyama:
The failure is because Sequel 5 was released and there's a compatibility issue. I'll peg the version and get the remaining specs to pass.

shioyama:
Cherry-picked this in cf3860f7edc0360f938f8ad35e6acbd648e4d0d3. Thanks!

shioyama:
Just FYI: I released [0.2.3](https://rubygems.org/gems/mobility/versions/0.2.3) which only has this fix, and created a new branch [0-2-stable:](https://github.com/shioyama/mobility/tree/0-2-stable) to track any further minor bug fixes prior to releasing 0.3.0.

Issue #51, opened 2017-07-08
Query with fallbacks


shioyama:
i.e.

```ruby
Mobility.locale = :en
Post.i18n.where(title: "foo")
```

...should return the post with title "foo" in English if a post with title in English exists, if not it should find a post with title "foo" in one of the fallback locales defined for the `title` attributes.

Ref: shioyama/friendly_id-mobility#4

shioyama:
So I've looked into this a bit today, and I think before doing anything like this it might be good to take a few steps back and think about how querying is currently implemented.

Currently, the locale is always implicit when querying, so `Post.i18n.where(title: "foo")` will implicitly look for posts whose title in English is `"foo"`.

I'm thinking that it would be useful to allow options ("query options") to be passed into the query method, which could e.g. include which locales to match for:

```ruby
Post.i18n(locales: [:en, :ja]).where(title: "foo")
#=> finds posts whose title in English or Japanese is "foo"
```

This seems like a logical step toward something like fallbacks, which could be a different option.

Where I'm hesitating a bit is whether to couple accessor fallbacks (the fallbacks plugin) to querying fallbacks. Globalize does this but I'm thinking of keeping them separate and isolated, but open to feedback.

shioyama:
I'm leaning now toward implementing this with a simple `locale` key in the query hash, like this:

```ruby
Post.i18n.where(title: "foo", locale: [:en, :fr])
```

One advantage here is that you can apply locales to specific attributes, allowing you to potentially build up more complex multi-locale queries:

```ruby
Post.i18n.where(title: "foo", locale: [:en, :fr]).where(content: "bar", locale: :de)
```

This is pretty far-fetched, but all else equal I think localizing the scope of the locale is better and this shouldn't be any more complicated than passing the locale to `i18n`.

On the other hand, this means that `locale` becomes a reserved key inside the `i18n` scope. I think this is fine, it pretty much *is* a reserved key anyway.

shioyama:
So #233 now implements passing the locale in the options to where. It only accepts a single locale, but accepting an array of locales should not be so hard given that the entire querying interface has been rewritten.

hellekin:
It looks to me like fallbacks do not work with `:container` backend:

``` ruby
[10] pry(main)> Mobility.config
=> #<Mobility::Configuration:0x000055bc07a2a468
 @accessor_method=:translates,
[...]

PR #86, opened 2017-09-08
Disable fallbacks when using locale/fallthrough accessors

### LINKED ISSUE #85: locale_accessors & fallbacks
When using locale_accessors, the locale specific accessors also use the fallbacks. e.g.,

``` ruby
Event.new(title_en: "foo").title_ja 
#=> "foo" instead of nil
```

I'm using the locale accessors to generate a form that provides the ability to edit an events title in Japanese and English (in two different form fields). With globalize, the locale specific accessors would return nil. This made it obvious which fields have not been translated. It also meant translations only got written when the user entered them. With mobility's current behaviour, if a user edits an event with an untranslated title and then saves it, the untranslated title will become the original title.

shioyama:
Fixes #85

### DIFF
--- a/spec/mobility/plugins/active_model/dirty_spec.rb
+++ b/spec/mobility/plugins/active_model/dirty_spec.rb
@@ -284,7 +284,7 @@ describe "Mobility::Plugins::ActiveModel::Dirty", orm: :active_record do
         expect(article.title).to eq("Title")
         expect(article.changed?).to eq(true)
         expect(article.changed).to match_array(["title_ja", "title_en"])
-        expect(article.changes).to eq({ "title_ja" => [nil, "ばばば"], "title_en" => ["ばばば", "Title"]})
+        expect(article.changes).to eq({ "title_ja" => [nil, "ばばば"], "title_en" => [nil, "Title"]})
       end
     end
   end
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ locations, usually database columns. By default these values are stored as keys
 tables, one for strings and one for text columns, but this can be easily
 changed and/or customized (see the [Backends](#backends) section below).
 
-### Getting and Setting Translations
+### <a name="getset"></a> Getting and Setting Translations
 
 The easiest way to get or set a translation is to use the getter and setter
 methods described above (`word.name` and `word.name=`), but you may want to
@@ -309,6 +309,11 @@ word.name(locale: :fr)
 #=> "mobilité"
 ```
 
+Note that setting the locale this way will pass an option `locale: true` to the
+backend and all plugins. Plugins may use this option to change their behavior
+(passing the locale explicitly this way, for example, disables
+[fallbacks](#fallbacks), see below for details).
+
 You can also *set* the value of an attribute this way; however, since the
 `word.name = <value>` syntax does not accept any options, the only way to do this is to
 use `send` (this is included mostly for consistency):
@@ -387,8 +392,8 @@ translated attributes on a class:
 ```ruby
 class Word < ApplicationRecord
   extend Mobility
-  translates :name,    type: :string, fallbacks: { de: :ja, fr: :ja }
-  translates :meaning, type: :text,   fallbacks: { de: :ja, fr: :ja }
+  translates :name,    type: :string, fallbacks: { de: :ja, fr: :ja }, locale_accessors: true
+  translates :meaning, type: :text,   fallbacks: { de: :ja, fr: :ja }, locale_accessors: true
 end
 ```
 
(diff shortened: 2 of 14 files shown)


spec/mobility/plugins/active_model/dirty_spec.rb
```diff
@@ -284,7 +284,7 @@ def save
         expect(article.title).to eq("Title")
         expect(article.changed?).to eq(true)
         expect(article.changed).to match_array(["title_ja", "title_en"])
-        expect(article.changes).to eq({ "title_ja" => [nil, "ばばば"], "title_en" => ["ばばば", "Title"]})
+        expect(article.changes).to eq({ "title_ja" => [nil, "ばばば"], "title_en" => [nil, "Title"]})
```
shioyama: As a byproduct, this change also makes dirty/fallbacks compatibility a bit less weird.

shioyama:
@pwim see what you think.

pwim:
The downside to this approach is that it creates a dependency between the locale_accessors and fallbacks plugins.

locale_accessors needs to know about the fallbacks plugin because the fallbacks plugin removes the ability to ask for a fallback in a given language

``` ruby
Mobility.locale = :ja
word = Word.create(name: "モビリティ", meaning: "(名詞):動きやすさ、可動性")
word.name(locale: :de)
#=> "モビリティ"
```

Personally, I think the following behaviour would be cleaner:

``` ruby
Mobility.locale = :ja
word = Word.create(name: "モビリティ", meaning: "(名詞):動きやすさ、可動性")
word.name(locale: :de)
#=> nil
Mobility.locale = :de
word.name
#=> "モビリティ"
```

i.e., the fallbacks would only be used when the locale is implied by the global one, not explicit. If this was the case, then it would be possible to implement the accessors in a way that doesn't explicitly couple it to the fallback plugin.

shioyama:
Ah, interesting. You're right that passing `locale: :de` should not use fallbacks either, so handling it the way you proposed would kill two birds with one stone, while removing coupling.

shioyama:
So this is the way I can envision implementing this. Currently, whether the locale is passed into a reader or writer with `locale: :de` or through the global `Mobility.locale`, the backend sees them exactly the same. We could change this so that in the reader method, instead of simply defaulting to the global locale, we set some kind of flag which the backend or plugins can use to optionally act differently (as in the case of locale/fallthrough accessors).

Something like this:

```ruby
define_method attribute do |**options|
  return super() if options[:super]
  locale = options[:locale] || Mobility.locale
  Mobility.enforce_available_locales!(locale)
  options[:locale] &&= !!locale
  mobility_backend_for(attribute).read(locale.to_sym, options)
end
```

So ignoring the `super` stuff which is unrelated: if a locale is passed in, then this is converted to a boolean and backends/plugins will see `locale: true` which they can interpret as meaning that the locale passed in as the first argument was explicitly set.

The changes to locale/fallthrough accessors in this PR can then be reverted since we don't need them, and instead *fallbacks* has to look for `locale: true` and in this case treat it as `fallback: false`. So basically rather than coupling locale accessors to fallbacks, we're instead coupling fallbacks to the core `Attributes` class, which is better since anyway they are already coupled.

I think using the `locale` key makes sense since we anyway already use it so it is "reserved" as a namespace on the accessor options hash.

pwim:
This approach sounds good to me. I think setting `locale: true` should work, and if I understand things correctly, it is only exposed to plugins anyways.

shioyama:
@pwim I've merged this into master, give it a try and let me know what you think. I'm debating whether to release it as 0.2.3 or combine it with other changes in the next minor version 0.3...

pwim:
Awesome! After pointing my Gemfile to point to master, applying the workaround described in #84, and working around two differences with globalize, I've switched Doorkeeper to use mobility. It's live now, with no immediately obvious issues. 

 Seeing as this is making an incompatible API change, [semantic versioning suggests we should release a new minor version](http://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase).

----

By the way, I discovered two slight differences with globalize:

1. Globalize's dirty handling marks the localized attribute as dirty (eg. `title`), whereas Mobility marks the locale specific attribute as dirty (eg. `title_ja`).
2. When you set a translation to `""` with Globalize, the localized attribute will return nil. For example `Event.new(title_ja: "").title #=>  nil`, whereas Mobility will return the empty string.

I think eventually you'll want to make a guide for migrating from Globalize to Mobility, so I wanted to report these as they might be worth noting. Though, they might be too specific to bother with, and won't make a difference for most people.

shioyama:
@pwim Thanks for investigating! And glad to hear you're using Mobility in production! That's great news.

About the two differences, the first one is a conscious decision, but the second one should not happen as long as you have the "presence" plugin enabled (it's enabled by default).

```ruby
Post.new(title_ja: "").title
=> nil
```

This is what I get with a model like this:

```ruby
  extend Mobility
  translates :title, backend: :table, locale_accessors: true
end
```

Did you set `default_options`? If you did and did not include `presence: true` then that would explain the behaviour you see.

shioyama:
About versions, I'll release cf3860f7edc0360f938f8ad35e6acbd648e4d0d3 as 0.2.3 and point that to a new branch `0.2`, and change the version on master to 0.3.alpha`. I'm going to wait for a few more things (including #84 and possibly an implementation for #90) before releasing 0.3.0.

shioyama:
The other thing missing from Globalize compatibility (which is harder to implement) is #51.

PR #93, opened 2017-09-15
Test integration with AR's update_attribute method

pwim:
Appears to be broken in the key_value/table backends. Works under Rails 5.0 if dirty plugin is enabled. Does not work even if the plugin is enabled under Rails 5.1.

Looking at update_column itself, it seems to rely on dirty handling, only saving if there are changes, which explains why it doesn't work without the dirty plugin in Rails 5.0. It's not so obvious to me why it would be broken under 5.1, and this possibly means there's something off with the dirty plugin under 5.1.

### DIFF
--- a/spec/mobility/backends/active_record/column_spec.rb
+++ b/spec/mobility/backends/active_record/column_spec.rb
@@ -109,5 +109,9 @@ describe "Mobility::Backends::ActiveRecord::Column", orm: :active_record do
       include_querying_examples 'Comment', :content, :author
       include_validation_examples 'Comment', :content, :author
     end
+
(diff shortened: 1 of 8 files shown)


shioyama:
5.1 issue may be related to this: https://github.com/rails/rails/pull/27963#issuecomment-310092787

See #44. I'll have a look later.

shioyama:
Can you also create an issue for this?

shioyama:
Can I close this, or do you want to leave a pending spec to watch for when this gets fixed in Rails?

pwim:
Good point, yeah I don't think this is need anymore.

PR #103, opened 2017-09-30
Ensure cache_key is invalidated when updating translations

### LINKED ISSUE #101: Updating translation doesn't bust the cache_key
When you update a translation (table backend), the cache key is not being invalidated. 

There's two approaches I see of fixing this:

#### Add `touch: true` when setting up the `belongs_to` relationship. 

The advantage of this approach is it is quite simple to implement, and makes the model behave like any of the backends that store the translation within the model.

The disadvantage of this approach is that if someone is using the `updated_at` for something more than cache invalidation, this might not match their expectation. 

Globalize has `touch` as an option. I've made [a branch of mobility using this approach](https://github.com/pwim/mobility/tree/table-touch), which I'll use for now on my app.

#### Overwrite the cache_key

This is the approach [that globalize takes](https://github.com/globalize/globalize/blob/ef1e362f3144b2276d89f47407c67ffa3fc1f9f2/lib/globalize/active_record/instance_methods.rb#L164-L166). 

One possible issue I see with their implementation is that it constructs the `cache_key` using only the translation based on the current locale. If you're only using methods like `event.title`, this works fine. But if you access a translation in a different locale than the one you're in (e.g., when the locale is `:en`, accessing `event.title(:ja)`), the cache_key will be incorrect.

More generally, I see this approach as potentially having performance implications, as now you need to load the translations to calculate the cache_key.

pwim:
For the table and key_value backends, this means the model itself is touched when updating a translation.

Fixes #101.

### DIFF
--- a/lib/mobility/active_record/translation.rb
+++ b/lib/mobility/active_record/translation.rb
@@ -4,7 +4,7 @@ module Mobility
     class Translation < ::ActiveRecord::Base
       self.abstract_class = true
 
-      belongs_to :translatable, polymorphic: true
+      belongs_to :translatable, polymorphic: true, touch: true
 
       validates :key, presence: true, uniqueness: { scope: [:translatable_id, :translatable_type, :locale] }
       validates :translatable, presence: true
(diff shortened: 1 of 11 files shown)


shioyama:
:+1: This is great, I'll merge once specs pass. Thanks!

shioyama:
Hmm seems you need to update some dirty specs since you've added timestamp columns that change. Seems to only affect specs in AR 4.2 and 5.0 for some reason.

Also with mysql in 5.1 doesn't seem to work... strange.

pwim:
Closed in favour of #104.

PR #111, opened 2017-11-27
Support added AR::Dirty methods

### LINKED ISSUE #106: Support Rails 5's `saved_changes`
Rails 5 introduced the [`saved_changes`](http://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Dirty.html#method-i-saved_changes) method to the Dirty module, but it seems like `Mobility` still doesn't support that:

```ruby
class Post
  include Mobility
  translates :translated_attribute, dirty: true
end

post = Post.first
post.translated_attribute = "test"
post.not_a_translated_attribute = "test"

post.changes 
# => { "translated_attribute" => ["", "test"], "not_a_translated_attribute" => ["", test] }

post.save!

post.saved_changed 
# => { "not_a_translated_attribute" => ["", "test"] }
```

It would be great if `saved_changes` also returned the translated attributes.

Thank you.

shioyama:
Fixes #106

### DIFF
--- a/lib/mobility/plugins/active_record/dirty.rb
+++ b/lib/mobility/plugins/active_record/dirty.rb
@@ -36,6 +36,10 @@ details on usage.
             end
 
             if ::ActiveRecord::VERSION::STRING >= '5.1'
+              define_method :saved_changes do
+                (@previously_changed ||= ActiveSupport::HashWithIndifferentAccess.new).merge(super())
+              end
+
               attribute_names.each do |name|
                 define_method :"saved_change_to_#{name}?" do
                   previous_changes.include?(Mobility.normalize_locale_accessor(name))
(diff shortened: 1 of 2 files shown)


shioyama:
@brianstorti Here's a PR which adds `saved_changes` as well as other AR::Dirty attribute methods. I'm a bit concerned about the transition to AR 5.2, since dirty specs are currently already all failing, but in any case the specs added here should be useful, even if the code has to change in the next version.

Have a try and see what you think.

shioyama:
So good news is that the changes in AR 5.2 don't seem to be related to the added methods here. Bad news is that I haven't got a good solution for them yet... but I'll merge this PR later today since it seems to work fine, and I don't think it will add any issues to fixing the 5.2 issues.

PR #116, opened 2017-11-29
Fix AM/AR Dirty plugin issues with AR 5.2

### LINKED ISSUE #115: Fix AM/AR::Dirty plugin compatibility issues with AR 5.2
## Context
ActiveRecord 5.2.beta.2 [has just been released](https://rubygems.org/gems/activerecord/versions/5.2.0.beta2), and we are now testing against it (see #112). There are about twenty specs failing currently, mostly around the Dirty plugin.

The main issue seems to be with rails/rails#28661 (fixing rails/rails#28660), which replaced a `__send__(attr)` to `_attributes(attr)`, and ultimately to `_read_attribute(attr)`. This is problematic since while in ActiveModel `_read_attribute` simply maps to `__send__`, which for a localized accessor like `title_en` will correctly return the current value of the title, in AR [it maps to](https://github.com/rails/rails/blob/95b86e57a6c07bc636f8feab4afd4c7fe0ca512c/activerecord/lib/active_record/attribute_methods/read.rb#L72-L74) this:

```ruby
def _read_attribute(attr_name, &block) # :nodoc
  @attributes.fetch_value(attr_name.to_s, &block)
end
```

Here, `@attributes` will not contain the localized translated attributes, so `title_en` here will return `nil`. So most of the AR::Dirty specs are failing with changes all coming out as `[nil, nil]` instead of e.g. `[nil, "foo"]`.

## Possible Fix

I'm currently investigating if we could override `_read_attribute` to do something special for translated attributes, but it's tricky because in the dirty plugin we're using localized attributes to track changes in multiple locales. This is tricky since without a complex regex, we can't really tell if a given attribute passed to `_read_attribute` is actually a translated attribute or not.

I don't want to be matching regexes every time `_read_attribute` is called since this is a *very* frequently-used method and regexes are slow, so this could significantly impact performance. For now just tracking the problem here, I'll see if there are other ways to fix the problem.

shioyama:
Fixes #115

### DIFF
--- a/lib/mobility/plugins/active_model/dirty.rb
+++ b/lib/mobility/plugins/active_model/dirty.rb
@@ -32,6 +32,7 @@ value of the translated attribute if passed to it.
           if model.changed_attributes.has_key?(locale_accessor) && model.changed_attributes[locale_accessor] == value
             model.send(:attributes_changed_by_setter).except!(locale_accessor)
           elsif read(locale, options.merge(fallback: false)) != value
+            model.send(:mobility_changed_attributes) << locale_accessor
             model.send(:attribute_will_change!, locale_accessor)
           end
           super
@@ -63,6 +64,10 @@ value of the translated attribute if passed to it.
             private :restore_attribute!
           end
 
+          def included(model_class)
+            model_class.include ChangedAttributes
+          end
+
           private
 
           # Get method suffixes. Creating an object just to get the list of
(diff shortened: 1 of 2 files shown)


shioyama:
@pwim So I found a way to fix the problem while maintaining the current interface. Basically, I realized that any time we need to check if an attribute is a Mobility attribute in the context of attribute changes, we can do this quickly and easily *if we track which attributes have changed by Mobility*.

So I add a method `mobility_changed_attributes` which returns a set, and before calling `attribute_will_change!`, just add the attribute name to the set. Then we can override `_read_attribute` as before, except that rather than a tricky regex, we just check if this set contains the attribute, if it does then dispatch to `__send__`.

This is not beautiful, but it fixes all (currenty) dirty issues in AR 5.2, should have minimal performance impact (also since it uses `def` instead of `define_method`, and only includes the `ReadAttribute` method once), and of course only impacts users who use the dirty plugin.

I tried to think about how we could do things differently without fallthrough accessors, but the issue is that there is a case where the user (say) does not use locale accessors, and does not specify `available_locales`. In this case, we really don't know what locales we should be matching against. Fallthrough accessors allow this case to be handled seamlessly, although at a performance cost (which I will mention in the readme).

Also, regardless, any solution I can imagine will have to override `_read_attribute`, given how AR now overrides AM::Dirty methods. The only way around this would be to explicitly add attributes to the `@attributes` hash, which seems like it would be asking for yet more problems.

So... if specs pass (locally only remaining failures are with the generators), I'd like to merge this and release in 0.3.0 with the changes, so we're (mostly) prepared for the Rails 5.2 launch.

shioyama:
Generator spec failures are from an issue in the specs, not the generators themselves. So basically this PR should fix all known issues with Rails 5.2.

pwim:
Cool. Took a look at it, and nothing jumps out at me.

PR #118, opened 2017-11-30
Fix generator spec issues with latest AR version

shioyama:
The generator specs currently use the `RAILS_VERSION` env variable, but we use the string `"latest"` to run specs against the latest versions of AR and Sequel, so those specs are currently failing even though they should actually pass. This fixes that issue.


shioyama:
I rebased this on top of #116. Tests are now running against AR 5.2 without `allow_failures`.

PR #120, opened 2017-11-30
Disable AR::Dirty method overrides for AR >= 5.2

shioyama:
These AM::Dirty methods are no longer overridden in AR::Dirty 5.2, so we don't need the overrides anymore.

### DIFF
--- a/lib/mobility/plugins/active_record/dirty.rb
+++ b/lib/mobility/plugins/active_record/dirty.rb
@@ -30,7 +30,7 @@ AR::Dirty plugin adds support for the following persistence-specific methods
           def initialize(*attribute_names)
             super
             @attribute_names = attribute_names
-            define_method_overrides
+            define_method_overrides if ::ActiveRecord::VERSION::STRING < '5.2'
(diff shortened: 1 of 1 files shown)


lib/mobility/plugins/active_record/dirty.rb
```diff
@@ -30,32 +30,39 @@ class MethodsBuilder < ActiveModel::Dirty::MethodsBuilder
...
+
+            if ::ActiveRecord::VERSION::MAJOR == 5 && ::ActiveRecord::VERSION::MINOR == 1
+              names = @attribute_names
+              method_name_regex = /\A(#{names.join('|'.freeze)})_([a-z]{2}(_[a-z]{2})?)(=?|\??)\z/.freeze
+              has_attribute = Module.new do
+                define_method :has_attribute? do |attr_name|
+                  super(attr_name) || !!method_name_regex.match(attr_name)
+                end
               end
+              model_class.extend has_attribute
```
shioyama: For once in a while I'm actually pleasantly surprised at changes in Rails. All this nasty stuff here we will be able to remove once we deprecate support for AR 5.1.

PR #121, opened 2017-12-01
Fix Sequel 5 issues

### LINKED ISSUE #94: Support Sequel >= 5.0
There are some specs failing when running against [Sequel 5.0](http://sequel.jeremyevans.net/2017/09/01/sequel-500-released.html) that need to be fixed. We can release Sequel 5.0 compatibility in 0.3.0.

shioyama:
Fixes #94

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,9 @@ env:
   - DB=postgres ORM=sequel SEQUEL_VERSION=4.46
   - DB=mysql ORM=sequel SEQUEL_VERSION=4.46
   - DB=sqlite3 ORM=sequel SEQUEL_VERSION=4.46
-  - DB=postgres ORM=sequel SEQUEL_VERSION=latest
-  - DB=mysql ORM=sequel SEQUEL_VERSION=latest
-  - DB=sqlite3 ORM=sequel SEQUEL_VERSION=latest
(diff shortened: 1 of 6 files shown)


shioyama:
So the good news is, Sequel 5 is mostly working fine. The only thing that is failing is query support for the Table backend, because there is code which sets an instance variable on a dataset, and this is no longer possible since datasets in Sequel are now [frozen by default](https://github.com/jeremyevans/sequel/commit/1c69ba426285ead2aeb28d7b28b73c422dece2cd) and [cannot be unfrozen by simply duping them](https://github.com/jeremyevans/sequel/blob/84e3fc4e9f6b74d2d6503c226d17bb89984fa882/lib/sequel/dataset/misc.rb#L50-L52), because `dup` simply returns (the frozen) dataset.

The problem in Mobility is [here](https://github.com/shioyama/mobility/blob/95a41dc528eb85a378f36e30cbcadd250b3b34db/lib/mobility/backends/sequel/table/query_methods.rb#L22-L23), where we set an instance variable to avoid joining the same table more than once. This is admittedly not very nice code, but it serves an important purpose...

shioyama:
Travis was being stupid about a couple builds, which seemed to be stuck, so I just merged this. It's passing anyway. Full support for Sequel 5.0 - yay!

Issue #123, opened 2017-12-02
Dup does not dup translations (KeyValue backend)


shioyama:
#84 fixed this issue for the Table backend, but it is still broken for the KeyValue backend.

## Context

`dup` does not duplicate translations. Suppose we have:

```ruby
post = Post.new
post.title = "foo"
post.save
post.title
#=> "foo"
post_dup = post.dup
```

## Expected Behavior

The `title` of the dup'ed post should be `"foo"`:

```ruby
post_dup.title
#=> "foo"
```

## Actual Behavior

The `title` is nil, because the translations are not dup'ed:

```ruby
post_dup.title
#=> nil
```

There is already a spec for this which is [disabled](https://github.com/shioyama/mobility/blob/38d458f3ebc9eb18c3d29b4b5fa4a71ba1bcf4e2/spec/support/shared_examples/dup_examples.rb#L54-L58) for KeyValue backends (AR + Sequel), so if someone works on this, just remove that skip code and make the test pass.

shioyama:
Maybe helpful: https://stackoverflow.com/questions/26287699/rails-deep-duplicate-a-record-with-polymorphic-children

PR #125, opened 2017-12-02
Add spec checking that dup'ed model translations are different objects

shioyama:
Working on the KeyValue `dup` fix (for #123), I noticed that the specs added in #84 would pass even if translations were not actually copied but actually were just pointing to the same object. So I added a simple spec verifying that changing the value of the attribute on the original instance should not change the value of the attribute on the dup'ed instance.

cc @pwim


shioyama:
Hmm ok, it seems Sequel does this for all backends... I guess I'll close this. I have a fix for AR KeyValue which passes this spec, I'll make a separate PR.

PR #126, opened 2017-12-03
Fix duping for AR KeyValue backend

### LINKED ISSUE #123: Dup does not dup translations (KeyValue backend)

shioyama:
Fixes #123

### DIFF
--- a/lib/mobility/backends/active_record/key_value.rb
+++ b/lib/mobility/backends/active_record/key_value.rb
@@ -70,6 +70,19 @@ Implements the {Mobility::Backends::KeyValue} backend for ActiveRecord models.
         end
         after_destroy :mobility_destroy_key_value_translations
 
+        module_name = "MobilityArKeyValue#{association_name.to_s.camelcase}"
+        unless const_defined?(module_name)
+          callback_methods = Module.new do
+            define_method :initialize_dup do |source|
+              super(source)
+              self.send("#{association_name}=", source.send(association_name).map(&:dup))
+              # Set inverse on associations
+              send(association_name).each { |translation| translation.translatable = self }
+            end
+          end
+          include const_set(module_name, callback_methods)
+        end
+
         private
 
         # Clean up *all* leftover translations of this model, only once.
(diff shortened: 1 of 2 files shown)


lib/mobility/backends/active_record/key_value.rb
```diff
@@ -70,6 +70,19 @@ def self.configure(options)
...
         after_destroy :mobility_destroy_key_value_translations
 
+        module_name = "MobilityArKeyValue#{association_name.to_s.camelcase}"
+        unless const_defined?(module_name)
+          callback_methods = Module.new do
+            define_method :initialize_dup do |source|
+              super(source)
+              self.send("#{association_name}=", source.send(association_name).map(&:dup))
+              # Set inverse on associations
+              send(association_name).each { |translation| translation.translatable = self }
```
shioyama: If I leave out the spec checking that changing the attribute on one instance does not affect the value on the other instance, then this can be simplified to just:

```ruby
define_method :initialize_dup do |source|
  super(source)
  self.send("#{association_name}=", source.send(association_name))
end
```

i.e. you don't need to dup the association objects, nor set the inverse.

However, I personally think that it's better that the dup'ed instance does not affect the original and vice versa. However, Sequel (for example) does not "deep dup" for any pattern here used in the backends... I suppose because there's a performance cost, and Sequel always defaults to not incurring any performance hits.

shioyama: Never mind, we definitely want duplicate models to have their own translations with AR at least. I think this is the correct way to fix this.

PR #136, opened 2017-12-08
Fix readme link to fallbacks rubydoc

madis:
Seems the fallbacks had moved to `Mobility::Plugins`. Updated the link
to correct rubydoc page url.


shioyama:
Wow, thanks very much! :+1: Totally missed those links, thanks for catching that.

PR #137, opened 2017-12-08
Make default plugin handle Procs more gracefully

shioyama:
Currently, you can use the `default` plugin with a `Proc`, with the proc being passed the model, attribute name, locale and options as keyword arguments. This ends up being quite clumsy to handle in practice.

This PR changes things so that instead, the context in the proc becomes the model, and the proc is passed up to three arguments: attribute, locale, options, but depending on the proc's arity it will only be passed enough arguments so that it can handle them.

This way, you can e.g. do this:

```
  extend Mobility
  translates :title, default: lambda { |attr| self[attr] }
end
```

...and this will default to reading the `title` column if the `title` attribute is not defined on the backend (`self[attr]` is equivalent to `read_attribute(attr)`).

### DIFF
--- a/lib/mobility/plugins/default.rb
+++ b/lib/mobility/plugins/default.rb
@@ -5,11 +5,12 @@ module Mobility
 Defines value or proc to fall through to if return value from getter would
 otherwise be nil.
 
-If default is a proc, it is passed a hash with four keyword arguments:
-- +model+: the model instance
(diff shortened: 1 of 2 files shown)


shioyama:
I decided to merge this, I'll add a deprecation and release another patch version. As of 0.4 this will be the API for this plugin.

Issue #81, opened 2017-08-31
default_fallbacks configuration option doesn't work (AR, Postgres)


smaximov:
Setting `default_fallbacks` configuration option doesn't have any (visible) effect. Consider the following snippet:

``` ruby
begin
  require 'bundler/inline'
rescue LoadError => e
  $stderr.puts 'Bundler version 1.10 or later is required.'
  raise e
end

gemfile(true) do
  source 'https://rubygems.org'

  gem 'activerecord', '5.1.0'
  gem 'mobility', '0.2.2'
  gem 'pg', '0.18.4'
  gem 'rspec', '3.6.0', require: 'rspec/autorun'
end

DB = ENV.fetch('MOBILITY_TEST_DB', 'mobility-test').freeze

ActiveRecord::Base.establish_connection(adapter: 'postgresql', database: DB)
ActiveRecord::Base.logger = Logger.new($stdout)

I18n.default_locale = :en
I18n.available_locales = %i[en ru]

Mobility.configure do |config|
  config.default_backend = :jsonb
  config.accessor_method = :translates
  config.default_fallbacks = { ru: :en, en: :ru }
end

ActiveRecord::Schema.define do
  create_table :articles, force: true do |t|
    t.jsonb :title, null: false, default: {}
  end
end

class Article < ActiveRecord::Base
  extend Mobility

  translates :title
end

RSpec.describe 'Fallbacks' do
  around do |example|
    I18n.with_locale(:en) { Article.create!(title: 'Title') }
    I18n.with_locale(:ru) { example.run }
  end

  subject(:article) { Article.last }

  it { expect(article.title).to eq('Title') }
end
```

I expected that setting `default_fallbacks` will make all translated attributes to fallback to existing locales by default. Moreover, when I try to use both `default_fallbacks` and the `fallbacks` option to `translates`, I get "`default_fallbacks`: undefined method `call` for {:ru=>:en, :en=>:ru}:Hash (NoMethodError)":

``` ruby
begin
  require 'bundler/inline'
rescue LoadError => e
  $stderr.puts 'Bundler version 1.10 or later is required.'
  raise e
end

gemfile(true) do
  source 'https://rubygems.org'

  gem 'activerecord', '5.1.0'
  gem 'mobility', '0.2.2'
  gem 'pg', '0.18.4'
end

DB = ENV.fetch('MOBILITY_TEST_DB', 'mobility-test').freeze

[...]

PR #148, opened 2017-12-22
Rename default_fallbacks configuration to new_fallbacks / fallbacks_generator=

shioyama:
See #81 and [this SO question](https://stackoverflow.com/questions/46547719/how-to-set-default-fallbacks-for-mobility).

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,7 +423,7 @@ end
 
 Internally, Mobility assigns the fallbacks hash to an instance of
 `I18n::Locale::Fallbacks.new` (this can be customized by setting the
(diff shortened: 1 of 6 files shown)


shioyama:
This is not enough on its own, but at least it should be clearer that the thing being set is not the fallbacks themselves. Once this has been released in 0.4 we can make a further change to clean this up for 1.0.

shioyama:
Note to self: Need to think some more about the naming and implementation of this configuration setting. The reader and writer methods do different things (reader returns an instance, writer sets a proc) which is confusing.

PR #157, opened 2018-01-19
Add new jsonb container backend

shioyama:
This is the first new backend to be added to Mobility since all the original ones were introduced in v0.1.x. I was inspired to add it looking at the Elixir hex package named [Trans](https://hex.pm/packages/trans), which stores all translations for a model in a single depth-2 jsonb column (`translations`). I've seen this also discussed [elsewhere](https://github.com/denkGroot/Spina/issues/266#issuecomment-312824064), and thought it would be interesting to implement.

This implementation piggy-backs on parts of the existing Jsonb backend, but in other parts is entirely independent. Specs seem to be passing.

Regarding naming, I'm using the term "container" since this is what Trans uses. In the code I've called the backend "JsonbContainer", but this is quite a mouthful so I'm thinking of just renaming it "container", so you'd use it like this:

```ruby
class Post
  extend Mobility
  translates :title, backend: :container
end
```

... or of course just set `config.default_backend = :container`.

Curious if anyone has any thoughts on this (backend itself, implementation, naming, etc.)

### DIFF
--- a/lib/mobility/backends/active_record/container.rb
+++ b/lib/mobility/backends/active_record/container.rb
@@ -1,0 +1,116 @@
+module Mobility
+  module Backends
+=begin
+
+Implements the {Mobility::Backends::Container} backend for ActiveRecord models.
+
+=end
+    class ActiveRecord::Container
+      include ActiveRecord
+
+      require 'mobility/backends/active_record/container/query_methods'
+
+      # @return [Symbol] name of container column
+      attr_reader :column_name
+
+      # @!macro backend_constructor
+      # @option options [Symbol] column_name Name of container column
+      def initialize(model, attribute, options = {})
+        super
+        @column_name = options[:column_name]
+      end
+
+      # @!group Backend Accessors
+      #
+      # @note Translation may be a string, integer, boolean, hash or array
+      #   since value is stored on a JSON hash.
+      # @param [Symbol] locale Locale to read
+      # @param [Hash] options
+      # @return [String,Integer,Boolean] Value of translation
+      def read(locale, _ = nil)
+        model_translations(locale)[attribute]
+      end
+
+      # @note Translation may be a string, integer, boolean, hash or array
+      #   since value is stored on a JSON hash.
+      # @param [Symbol] locale Locale to write
+      # @param [String,Integer,Boolean] value Value to write
+      # @param [Hash] options
+      # @return [String,Integer,Boolean] Updated value
+      def write(locale, value, _ = nil)
+        set_attribute_translation(locale, value)
+        model_translations(locale)[attribute]
+      end
+      # @!endgroup
+
+      # @!group Backend Configuration
+      # @option options [Symbol] column_name (:translations) Name of column on which to store translations
+      def self.configure(options)
+        options[:column_name] ||= :translations
+      end
+      # @!endgroup
+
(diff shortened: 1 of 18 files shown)


shioyama:
Note to self: still need to implement for Sequel.

shioyama:
A query like `Post.i18n.where(title: "foo", content: "bar")` becomes:

```sql
SELECT "posts".* FROM "posts"
WHERE ("posts"."translations" -> 'en' @> '{"title":"foo"}'
AND "posts"."translations" -> 'en' @> '{"content":"bar"}')
```

This could actually be done with one query:

```sql
SELECT "posts".* FROM "posts"
WHERE ("posts"."translations" -> 'en' @> '{"title":"foo","content":"bar"}')
```

I'm not sure if there is any performance impact of querying on each key/value separately...

shioyama:
And here's a query with a nil value, `Post.i18n.where(title: "foo", content: nil)`, where we use the existence operator:

```sql
SELECT "posts".* FROM "posts"
  WHERE "posts"."translations" -> 'en' @> '{"title":"foo"}'
  AND NOT ("posts"."translations" ? 'en' AND "posts"."translations" -> 'en' ? 'content')
```

And a `not` query, negating the previous one: `Post.i18n.where.not(title: "foo", content: nil)` becomes:

```sql
SELECT "posts".* FROM "posts"
  WHERE NOT (1=0)
  AND "posts"."translations" ? 'en'
  AND "posts"."translations" -> 'en' ? 'title' AND NOT ("posts"."translations" -> 'en' @> '{"title":"foo"}')
  AND "posts"."translations" ? 'en' AND "posts"."translations" -> 'en' ? 'content'
```

We are thus looking for posts that have a `title` *other* than "foo", and which have a content which is non-nil.

Like the jsonb backend, we depend here on the fact that when storing data, we strip any values which are not present. This allows us to use the existence operator (`?`) without having to worry about keys with blank/null values -- we just assume there are none.

We can also store/query on non-string values (again, like the jsonb backend). e.g. `Post.i18n.where(title: {"a" => "b"})` would produce:

```sql
SELECT "posts".* FROM "posts"
WHERE "posts"."translations" -> 'en' @> '{"title":{"a":"b"}}'
```

And this will correctly match posts which have a hash value of `{"a": "b"}`. The value stored in the `translations` column in this case would look like this:

```ruby
{"en"=>{"title"=>{"a"=>"b"}}}
```

shioyama:
@pwim I'm curious is this is a backend you would consider using. It has a nice balance of ease-of-use (just one column per translated model), performance (jsonb so fast to query) and relatively low complexity (relative to, say, storing translations on a separate table).  Implementation is actually quite simple, even querying.

lib/mobility/backends/active_record/container.rb
```diff
@@ -0,0 +1,116 @@
...
+              class_eval <<-EOM, __FILE__, __LINE__ + 1
+                def initialize_dup(source)
+                  super
+                  self.#{column_name} = source.#{column_name}.deep_dup
+                end
+              EOM
+            end
+            include const_set(module_name, dupable)
+          end
+        end
```
shioyama: This is a lot of code just to fix an issue which is arguably a bug in AR 4.2... debating maybe just skipping the dup specs for this backend in that version of Rails.

PR #166, opened 2018-02-07
Update for AR 5.2.0.rc2

### LINKED ISSUE #164: Tests failing on Rails 5.2.0.rc1
They were passing on beta, but some seem to be failing now on rc1. Need to investigate.

shioyama:
Specs fail on 5.2.0.rc1, but some changes in that version have been [reverted](https://github.com/rails/rails/commit/5fcbdcfb574c731841be12764c50d9587b58345f), so I've updated the Gemfile to look at master and I'll try to fix the remaining issues here.

Against master at this time, six specs (related to dirty attribute tracking) are currently failing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,10 +12,8 @@ group :development, :test do
     elsif ENV['RAILS_VERSION'] == '5.1'
       gem 'activerecord', '>= 5.1', '< 5.2'
     else
-      git 'https://github.com/rails/rails.git', branch: '5-2-stable' do
-        gem 'activerecord'
-        gem 'railties'
-      end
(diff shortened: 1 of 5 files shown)


shioyama:
Ok this seems to fix failing specs for me locally, by partially reverting the change in rails/rails@e126078a0. However, since that commit is less than 10 hours old, and since this will have performance implications which that commit was intended to avoid, I'll wait to merge this until things "settle down".

In any case, if anyone is actually building with `--pre`, dirty may act funny, but it appears that Rails itself is in flux around this issue so need to wait a bit and see.

PR #182, opened 2018-03-14
Support postgres json type

shioyama:
Currently Mobility only  supports the `jsonb` column type. This PR adds support for the `json` type.


shioyama:
Fixes #182

PR #187, opened 2018-03-16
Array.wrap -> Kernel::Array

shioyama:
Why use ActiveSupport when you don't have to?


shioyama:
Actually they're [not exactly the same](https://stackoverflow.com/questions/18358717/ruby-elegantly-convert-variable-to-an-array-if-not-an-array-already). Tests should check whether the difference matters but I'll also double check relevant changes before merging this.

shioyama:
We need to keep `Array.wrap` in the pg query methods module because for jsonb columns, we want to be able to query on translated hash objects, but `Array(...)` will convert a hash by turning into a set of 2-element key/value arrays which is not what we want. So I left that one and converted the other two, which don't handle queries on hash-valued translations so are not affected.

PR #190, opened 2018-03-21
Refactor Container backend query methods

shioyama:
This changes the queries to use `IS NOT NULL` instead of using the jsonb-specific existence (`?`) operator. But I believe existence may be preferable for performance reasons. Need to research before merging this (or not merging) - if anyone has any info please let me know.


shioyama:
Looks like using existence (`?`) makes the query indexable, whereas `IS NOT NULL` is not. Also specs ssem to be failing in AR 4.2.

shioyama:
Ref: https://stackoverflow.com/questions/42502202/optimizing-postgres-jsonb-query-with-not-null-constraint

PR #195, opened 2018-03-23
Simplify Util module

shioyama:
We should only define the level of functionality that Mobility requires in this module, so e.g. presence should only check `nil` or `""`, etc.


shioyama:
Never mind, need to rethink this.

PR #198, opened 2018-03-25
Remove table backend index on foreign key alone

shioyama:
Since we already generate a unique index on the `[foreign_key, :locale]` pair, the index on `foreign_key` alone is unnecessary.

@pwim Curious about this, what indices do you have on your translation tables? See the change here. I don't think we need `:post_id` if we already have `[:post_id, :locale]`, and if we don't need an index better not to generate it.

### DIFF
--- a/lib/rails/generators/mobility/templates/table_translations.rb
+++ b/lib/rails/generators/mobility/templates/table_translations.rb
@@ -17,7 +17,6 @@ class <%= migration_class_name %> < <%= activerecord_migration_class %>
       t.timestamps null: false
     end
 
-    add_index :<%= table_name %>, :<%= foreign_key %>, name: :<%= translation_index_name(foreign_key) %>
     add_index :<%= table_name %>, :locale, name: :<%= translation_index_name("locale") %>
     add_index :<%= table_name %>, [:<%= foreign_key %>, :locale], name: :<%= translation_index_name(foreign_key, "locale") %>, unique: true
 
--- a/spec/generators/rails/mobility/translations_generator_spec.rb
+++ b/spec/generators/rails/mobility/translations_generator_spec.rb
@@ -70,7 +70,6 @@ describe Mobility::TranslationsGenerator, type: :generator, orm: :active_record 
                 contains "t.string  :locale, null: false"
                 contains "t.integer :post_id, null: false"
                 contains "t.timestamps null: false"
-                contains "add_index :post_translations, :post_id, name: :index_post_translations_on_post_id"
                 contains "add_index :post_translations, :locale, name: :index_post_translations_on_locale"
                 contains "add_index :post_translations, [:post_id, :locale], name: :index_post_translations_on_post_id_and_locale, unique: true"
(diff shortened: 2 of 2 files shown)


shioyama:
To be clear, when you run `rails generate mobility:translations Post title:string`, Mobility will currently generate three indices:

```ruby
add_index :post_translations, :post_id, name: :index_post_translations_on_post_id
add_index :post_translations, :locale, name: :index_post_translations_on_locale
add_index :post_translations, [:post_id, :locale], name: :index_post_translations_on_post_id_and_locale, unique: true
```

The first one seems redundant to me. For comparison, Globalize, generates just the first two, and if you include a `:unique_index` option it will generate the last one. But I can't really imagine any situation where you would *not* want a unique index in this situation (i.e. would you ever want multiple translations for the same locale? I can't imagine why). In Mobility this is default *on*, in which case I think the first index should be removed (what this PR does).

(Note you can add indexes on specific columns with `rails generate mobility:translations Post title:string:index` etc, which will add an index on `[:title, :locale]`.)

shioyama:
@pwim Thanks, that's very useful info.

> Given that globalize/mobility expect there only to be one translation per locale, I think not creating a unique index is a bug.

Yes I agree. Mobility is creating that index by default (rather than as an option) so no problem there.

Regarding removing the other index, my experience with MySQL is that if you have indexes for `[:a, :b]` and `:b`, then an index for `:a` alone is redundant, which is why I created this PR. I've also had situations where MySQL will pick the wrong index, actually resulting in worse (sometimes much worse) performance.

But the situation here is quite simple, so I don't think that should happen. Maybe I can just add a comment in the actual migration mentioning that that index can be removed? I honestly don't really want to go digging into the internals of every db's indexing mechanisms...

pwim:
Well since it is OK to remove for MySQL and Postgres, I supposed that covers most of the market share of databases that will be using it. I'd remove it then, as anyone who it becomes an issue for can easily add it after the fact.

shioyama:
Reading up, for MySQL at least, comments and docs seem to confirm that the index on `post_id` is redundant and since it takes up space (and time to create/update), better not to have.

e.g.

https://bytes.com/topic/mysql/answers/463047-indexed-single-col-search-capability-multi-col-index
> Indexes take up space on disk and in memory. Theres a "key cache" which
> holds index structures in memory. The size of the key cache is finite, and
> you can tune the allocation of memory for MySQL's key cache. A
> single-column index on A would naturally be smaller than the multi-column
> index on (A,B), so more of the index can fit in the same amount of memory.
> But if you intend to size the key cache so that the whole (A,B) index fits
> in it, then I'd say yes, the additional index is redundant.
> 
> Another consideration is whether the (A,B) index is redundant if you also
> have indexes on A and B individually. The best reason for keeing the (A,B)
> index is if uniqueness is enforced only over both columns instead of either
> A or B individually. Another reason to keep the (A,B) index is that MySQL
> has a limitation that only one index can be used per table per query. If
> you typically have searches that involve both A and B in one query, then an
> index over both columns could be beneficial.

We do want to enforce uniqueness, so (A, B) + (B) is the right choice here. I'll merge this.

Thanks @pwim :smile:

PR #199, opened 2018-03-25
Add prefix/suffix options to hash backends

### LINKED ISSUE #189: Add column prefix/suffix as options to json/jsonb/hstore backends
Related to #91, even if we do not default to a different column name in json/jsonb/hstore backends, there should at least be an option to add a prefix and/or suffix to the translated attribute name to map it to the column, to avoid issues like the ones mentioned in that issue.

shioyama:
Fixes #189


shioyama:
There's a lot of duplication in specs which should be removed, but for now I just want to get prefixes/suffixes working on all hash-valued backends in both ORM and release to allow users to use this feature asap.

PR #202, opened 2018-04-10
Add extension to find translations in a locale for Table backend

shioyama:
This may also be useful for KeyValue backend, but for now just adding it for the Table backend. Now you can call:

```ruby
post.translations.in(:en)
```

... and you get all translations in that locale. Currently, you'd need to do this:

```ruby
post.translations.find { |t| t.locale == "en" }
```

which is kind of annoying, especially if you do it frequently. I also refactored the backend to use this extension method.

### DIFF
--- a/lib/mobility/backends/active_record/table.rb
+++ b/lib/mobility/backends/active_record/table.rb
@@ -133,7 +133,8 @@ columns to that table.
           foreign_key: options[:foreign_key],
           dependent:   :destroy,
           autosave:    true,
-          inverse_of:  :translated_model
+          inverse_of:  :translated_model,
+          extend:      TranslationFinderExtension
 
         translation_class.belongs_to :translated_model,
           class_name:  name,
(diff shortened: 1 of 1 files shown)


shioyama:
I debated about naming, either `for` or `in`. Currently, `translation_for` on the backend returns a translation *or* builds a blank one if none exists. The extension method `in` here returns `nil` if none exist. We could have an option to build if one is missing like Globalize does... but I think, in the interest of simplicity, for now it will just take a locale and return a translation or `nil`.

shioyama:
@pwim would this change be useful to you? I found it useful fixing an issue in refinery/refinerycms#3360 Wondering about the method name. `in` is short and sweet.

shioyama:
Yes, I'd actually considered `in_locale`. I justified the choice of `in` to myself by the fact that this method is only applied to translations, not to the model or anything else. So if there was a conflict, it would either be with ActiveRecord itself, or with something the application was doing on translations already.

That said though, I could imagine AR using a method like `in` in the future...

shioyama:
Decided it's safer to go with `in_locale`, as suggested. Thanks @pwim for quick feedback, always nice to get your opinion on these things.

PR #210, opened 2018-04-18
[WIP] Allow passing locale to queries

### LINKED ISSUE #51: Query with fallbacks

shioyama:
Fixes #51.

This PR works for all backends except column currently. Ironically, column is the simplest backend but supporting an array-valued `locale` option to queries (`Post.i18n.where(title: "foo", locale: [:en, :fr])`) is more complicated than I thought.

I'm going to try getting basic specs green first, then think about how this can be cleaned up. I have a rough plan to lean much more heavily on Arel and hopefully make the query interface more flexible.

### DIFF
--- a/lib/mobility/backends/active_record/column/query_methods.rb
+++ b/lib/mobility/backends/active_record/column/query_methods.rb
@@ -4,12 +4,23 @@ require "mobility/backends/active_record/query_methods"
 module Mobility
   module Backends
     class ActiveRecord::Column::QueryMethods < ActiveRecord::QueryMethods
-      def initialize(attributes, _)
+      attr_reader :arel_table
+
+      def initialize(attributes, options)
         super
+        @arel_table = options[:model_class].arel_table
+
         q = self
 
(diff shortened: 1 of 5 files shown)


shioyama:
The pleasant surprise is that except for Column, other backends were quite easy to adapt to this feature.

shioyama:
Although specs are passing, this is not quite working yet. It breaks if you query on more than one translated attribute with different backends in a single `where`, because one backend strips the `locale` key/value so other backends don't get it.

Will work on a fix, but for basic cases this should work.

shioyama:
Closing this since #216 will require many changes. I'll make a new PR later once query stuff settles down a bit.

Issue #141, opened 2017-12-13
Breaks shoulda's validate_uniqueness_of matcher


jmuheim:
I'm testing uniqueness validations using shoulda like this:

```ruby
  describe 'uniqueness validations' do
    subject { build :project, creator: @user }

    it { should validate_uniqueness_of(:name).scoped_to(:customer) }
  end
```

Since I introduced mobility, this doesn't work for translated attributes anymore:

```
     Failure/Error: should validate_uniqueness_of(:name)
     
       Product did not properly validate that :name is case-sensitively unique.
         After taking the given Product, whose :name is ‹"Product test name"›,
         and saving it as the existing record, then making a new Product and
         setting its :name to a different value, ‹"pRODUCT TEST NAME"›, the
         matcher expected the new Product to be valid, but it was invalid
         instead, producing these validation errors:
     
         * name: ["has already been taken"]
```

When trying manually, the validation works as always. Only the shoulda matcher seems to have a problem.

jmuheim:
See https://github.com/jmuheim/base/blob/master/spec/models/page_spec.rb#L13

shioyama:
Yes, the uniqueness validation in Mobility does not handle case sensitivity, so this is an expected result. I should maybe document that somewhere.

The problem is that Mobility has to do validation in a backend-independent way, and doing a `LOWER` comparison requires that you know something about the backend strategy. e.g. it would be different for a jsonb/hstore column.

jmuheim:
Thanks for your response. So this means that Mobility replaces the default mechanism or what? And that replacement doesn't watch out for LOWER comparison?

IMHO, I think there should be at least an option to force LOWER comparison somewhere?

shioyama:
There is no mechanism for translated attributes, since they are not "real" attributes. Mobility adds a roughly-compatible uniqueness validation. I added it with some hesitation because I knew that it would not be complete.

I won't add any backend-specific conditions like `LOWER` because it makes things much more complciated. If it's important to have a case-sensitive comparison, then you can add a custom validator, or validate on the actual columns.

e.g. I believe you are using the column backend. In this case, you can validate uniqueness of the column, with the locale suffix. So you could `validate :title_en, uniqueness: true`, and it would validate the title in English.

I'm closing this since it's not fixable.

shioyama:
If you want to ask more about what you could do to handle this validation, please add a question on SO with the "mobility" tag.

shioyama:
[...]

PR #216, opened 2018-04-29
Restructure querying into plugin

shioyama:
This is far from complete, but I wanted to create a WIP right away to explain it since it's a big change with some benefits for anyone using Mobility with querying.

The change is to migrate the querying support offered by Mobility into a shared plugin. Rather than each backend independently overriding methods like `where`, the plugin does this overriding in one place and calls two methods on the backend *class*, which each backend can customize.

These methods are `build_node` and `add_scope` (exact names may change). Each backend only needs to implement these methods to support querying.

`build_node` takes an attribute name, a locale (currently `Mobility.locale`) and a set of backend options and returns an Arel node. `add_scope` takes a relation, a hash of attributes and values, and a hash of backend options and returns an (optionally) modified relation.

Why is this important? Because it means that we can fully abstract how backends handle querying into these two methods. e.g. for the Table backend (for which I've got specs now passing), `build_node` creates a node for the column on the translation table, whereas `add_scope` joins the translation table to the model table on the relation.

For the jsonb backend, the node would be e.g. `post.title -> 'en'`, for an English locale, etc.

This is really powerful because it means that we can now do more complex querying. e.g. suppose you want to case-sensitively search for a translated value independent of backend. With this change, if you get the node with `build_node`, you can then build a predicate with `node.lower.eq("foo")`. Previously, there was no backend-independent way of doing this.

This means that e.g. the [uniqueness validator](https://github.com/shioyama/mobility/blob/d09efcee30edb70ec69bf240c63a0cdb45471a04/lib/mobility/active_record/uniqueness_validator.rb) could support case-insensitivity, fixing #141. But it also means it would be possible to build extension gems to support integration with e.g. Ransack, etc. in a backend-independent way.

The code here is quite complex but when this is all done, things should be simpler since all the backend-specific `QueryMethods` classes will be removed, so complexity will be captured exclusively by one (plugin) class.


shioyama:
p.s. currently I'm focusing exclusively on ActiveRecord, but if/once I get this working I'll try to do a similar refactor for Sequel.

shioyama:
Specs are passing locally for Table, Jsonb, Json and Hstore backends. Container is just a bit more work, KeyValue will be slightly trickier but generally this seems to work well.

shioyama:
Ok this is working now locally for all AR backends. Still need to clean up a lot and support Sequel, but this chops off 300+ lines of code. :scissors:

shioyama:
やった! This is finally cleaned up and passing specs + code climate (after a few tweaks).

Chopped 250 lines and I think the code is cleaner but also more powerful. I'll need to add some docs and squash, etc., I figure the final line drop will be more in the ballpark of minus 200, which is still pretty good.

shioyama:
Rebased on top of #218

shioyama:
Rebased on #219.

PR #221, opened 2018-05-09
[WIP] Support case-sensitive uniqueness validation

shioyama:
This is a first attempt to leverage the new query plugin added in #216. The PG nodes were missing support for `LOWER` so I've added it.

Related: #141

### DIFF
--- a/lib/mobility/arel/nodes/pg_ops.rb
+++ b/lib/mobility/arel/nodes/pg_ops.rb
@@ -8,6 +8,7 @@ module Mobility
         JsonDashArrow
         JsonDashDoubleArrow
         JsonbDashArrow
+        JsonbDashDoubleArrow
         JsonbQuestion
         HstoreDashArrow
         HstoreQuestion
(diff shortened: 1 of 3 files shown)


Gbeschbacher:
Hi @shioyama, is this PR also related to do a uniqueness validation of an attribute with json backend? As an example, i've got an attribute and 2 locales and i want that the attribute is unique based on the locale scope.

shioyama:
@Gbeschbacher Yes, Mobility currently already supports uniqueness by locale for every backend, see [here](https://github.com/shioyama/mobility/wiki/Migrating-from-Globalize#uniqueness-validation). However it is case-sensitive uniqueness only.

I'm planning to support case-insensitive uniqueness as well once some big changes are done. Currently if you try to use case sensitive uniqueness Mobility will issue a warning telling you it is not supported.

Gbeschbacher:
Thanks for the very fast reploy @shioyama! I'm also already using MySQL in the same Setup as metioned  [here](https://github.com/shioyama/mobility/issues/226). Guess I'm having some problems with MySQL there not related to your gem.

```
Class Category
  extend Mobility
  translates :name
  validates :name, uniqueness: true, presence: true
```

`Category.find(1).update!(name: "foobar")` gives me an

```
ActiveRecord::StatementInvalid: Mysql2::Error: Invalid JSON path expression. The error is around character position 1.: SELECT  1 AS one FROM `categories` WHERE (`categories`.`name` ->> 'en') = 'foobar' AND `categories`.`id` != 1 LIMIT 1
from /usr/local/bundle/gems/mysql2-0.5.1/lib/mysql2/client.rb:131:in `_query'
```

shioyama:
Hmm, reading the MySQL docs now, looks like perhaps the syntax is actually a bit different: `c->>'$.name'`, which is similar but different to PostgreSQL. That would explain the error you see.

shioyama:
I'm closing this since before we can support case-insensitive uniqueness validation, we need to support `lower` on all backend nodes. Currently it is missing from json/jsonb/hstore/container backends.

PR #225, opened 2018-05-17
Use backend visitors to fully support Arel queries

shioyama:
(This is a follow-up to #216.)

This PR implements **full support** for querying using Arel translation nodes across all backends. It's taken a lot of thought to get here, but I think this will be a very important change for anyone using Mobility with querying, particularly complex querying.

In previous versions of Mobility, you could call `where` and `not` on a scoped relation on a translated model, and Mobility would convert translated attributes into their corresponding backend-specific nodes. However, this meant that you were limited to equality-type predicates, since that's all that `where` supports.

So I thought: what if you could just get an Arel node for a backend attribute? Then you could build your own queries with the node, e.g. `title.eq("foo")` but also `title.matches("foo")`, or anything else you could imagine.

The tricky part with this is that some backends (Table and KeyValue) have associations they depend on. A node on its own can't arbitrarily join another table, so the node would depend on backend-specific code (joining one or more tables). I want a truly backend-independent solution here, so this was not enough.

PR #216 added another method `add_translations` which had to be called on the backend to fully support querying on Arel nodes, but I didn't really like this implementation.

After digging further, I came to a better solution which instead adds visitor classes for Table and KeyValue bakends (like Arel uses internally). These visitors take a predicate, a relation and a locale, and return the relation with whatever the relation will need to support the predicate. The details are too complex here (I'll write about it in a blog post soon), but the result is that there is now a powerful interface to query on any backends with Arel nodes.

Here is an example with a deliberately complex model:

```ruby
  extend Mobility
  translates :title, backend: :jsonb
  translates :subtitle, backend: :table
  translates :author, backend: :column
end
```

The interface uses the existing `i18n` scope, but you can now optionally pass attribute names and a block, and the block will be passed the nodes from which you can build a predicate:

```ruby
Post.i18n(:title, :subtitle, :author) do |title, subtitle, author|
  title.eq("Querying with Arel in Mobility").
    and(subtitle.eq("Feel the Power")).
    and(author.matches("Chris"))
```

This will generate the SQL:

```sql
SELECT "posts".* FROM "posts"
  INNER JOIN "post_translations" ON "post_translations"."post_id" = "posts"."id" AND "post_translations"."locale" = 'en'
  WHERE ("posts"."title" -> 'en') = '"Querying with Arel in Mobility"'
  AND "post_translations"."subtitle" = 'Feel the Power'
  AND "posts"."author_en" ILIKE 'Chris'
```

Note that each attribute is mapped to a different predicate depending on the backend, but the predicate you build in the block is completely independent of these details. Also, the `JOIN` is added because the `subtitle` attribute needs it, but again you don't need to worry about.

Not only that, but Mobility will detect whether you need an `INNER` or `OUTER` join, so if I change the query to this:

```ruby
Post.i18n(:title, :subtitle, :author) do |title, subtitle, author|
  title.eq("Querying with Arel in Mobility").
    and(subtitle.eq(nil)).
    and(author.matches("Chris"))
```

then the query becomes:

```sql
SELECT "posts".* FROM "posts"
  LEFT OUTER JOIN "post_translations" ON "post_translations"."post_id" = "posts"."id" AND "post_translations"."locale" = 'en'
  WHERE ("posts"."title" -> 'en') = '"Querying with Arel in Mobility"'
  AND "post_translations"."subtitle" IS NULL
  AND "posts"."author_en" ILIKE 'Chris'
```

Note here that since we're looking for a `NULL` value on `post_translations`, we need to `OUTER JOIN` the table because the translation may be missing (which means the value is treated as `nil`). Without an `OUTER JOIN` here, you could return a miss when in fact there was a post it should have matched.

Using the [Visitor Pattern](http://patshaughnessy.net/2014/9/23/how-arel-converts-ruby-queries-into-sql-statements) here makes this query interface very powerful, since Mobility will evaluate the entire tree of the predicate, so even if you create more complex nodes it will still find the translated node(s) and join the correct table in the correct way.

### DIFF
--- a/lib/mobility/arel.rb
+++ b/lib/mobility/arel.rb
@@ -1,1 +1,18 @@
+# frozen-string-literal: true
 require "mobility/arel/nodes"
+require "mobility/arel/visitor"
+
+module Mobility
+  module Arel
+    class Attribute < ::Arel::Attributes::Attribute
+      attr_reader :backend_class
+      attr_reader :attribute_name
+
+      def initialize(relation, column_name, backend_class, attribute_name = nil)
+        @backend_class = backend_class
+        @attribute_name = attribute_name || column_name
+        super(relation, column_name)
+      end
+    end
+  end
+end
--- a/lib/mobility/arel/nodes.rb
+++ b/lib/mobility/arel/nodes.rb
@@ -4,8 +4,8 @@ module Mobility
     module Nodes
       class Unary  < ::Arel::Nodes::Unary;  end
       class Binary < ::Arel::Nodes::Binary; end
-      class Grouping < Unary; end
-      class Equality < Binary; end
+      class Grouping < ::Arel::Nodes::Grouping; end
+      class Equality < ::Arel::Nodes::Equality; end
 
       ::Arel::Visitors::ToSql.class_eval do
         alias :visit_Mobility_Arel_Nodes_Equality :visit_Arel_Nodes_Equality
--- a/lib/mobility/arel/visitor.rb
+++ b/lib/mobility/arel/visitor.rb
@@ -1,0 +1,61 @@
+# frozen-string-literal: true
+module Mobility
+  module Arel
+    class Visitor < ::Arel::Visitors::Visitor
+      INNER_JOIN = ::Arel::Nodes::InnerJoin
+      OUTER_JOIN = ::Arel::Nodes::OuterJoin
+
+      attr_reader :backend_class
+
+      def initialize(backend_class = nil)
+        super()
+        @backend_class = backend_class
+      end
+
+      private
+
+      def visit(object)
+        super
+      rescue TypeError
+        visit_default(object)
+      end
+
+      def visit_collection(_objects)
+        raise NotImplementedError
+      end
+      alias :visit_Array :visit_collection
+
+      def visit_Arel_Nodes_Unary(object)
+        visit(object.expr)
+      end
+
+      def visit_Arel_Nodes_Binary(object)
+        visit_collection([object.left, object.right])
+      end
+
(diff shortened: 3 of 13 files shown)


shioyama:
I simplified the format by using virtual rows, inspired by [Sequel](http://sequel.jeremyevans.net/rdoc/files/doc/virtual_rows_rdoc.html). So now you can do complex Arel queries without any boilerplate or any arguments at all:

```ruby
Post.i18n do
  title.eq("Querying with Arel in Mobility").
    and(subtitle.eq("Feel the Power")).
    and(author.matches("Chris"))
end
```

Bare words like `title`, `subtitle` and `author` are evaluated in a basic object and mapped to Mobility nodes if the name matches a translated attribute.

I actually wonder why ActiveRecord doesn't support this format with `where`, i.e. `Post.where { created_at.gt(1.day.ago) }`, seems like it would make Arel much more accessible.

shioyama:
So this now works with multiple tables, so if you have comments and posts, and they both have a translated `content` attribute, you can find all posts with content that matches a comment content like this:

```ruby
Post.joins(:comments).i18n{ |p| Comment.i18n{ |c| p.content.eq(c.content) } }
```

So the block now takes an optional argument, and if provided the virtual row (`p` and `c` above) is passed in rather than set on the context of the block. This way, you can combine nodes like I did here. The return value of the inner block here is a relation, so the outside block sees this and strips out the predicate using `relation.arel.constraints.last`. This is quite tricky so for now it'll be a bit experimental, but enables some very complex querying on translated attributes across different models/tables.

shioyama:
Finally merged! Will hopefully ship a new release with this next week.

Issue #170, opened 2018-02-11
Set no default type on KeyValue backend, remove from readme


shioyama:
We should remove the `type` option in examples in the readme since it only applies to the KeyValue backend, and force users to set a default on this option for KeyValue using `default_options`.

## Context

The readme currently includes `type: :string` etc in all examples, because the default backend is KeyValue which will default to `type: :text` otherwise. Other backends ignore the `type` option altogether, but because it's in the readme people get confused and think it applies to all backends.

## Possible Fix

With the KeyValue backend, if you have:

```ruby
class Post < ApplicationRecord
  translates :title
end
```

then by default `Post` will use the `mobility_text_translations` table which uses a text column to store the translation. To override this you need to pass `type: :string` to `translates.

What we should do instead is just *for the KeyValue backend* raise an error if `type` is not passed as an option, so that the default is no longer `text`. Setting `type: ...` everywhere may be cumbersome, so users can set:

```ruby
Mobility.configure do |config|
  # ..
  config.default_options[:type] = :string
  # ..
end
```

and this will default the `type` option to `:string` everywhere.

Once this is done, we can remove references to the `type` option everywhere in the readme, and avoid confusion for people using non-KeyValue backends.

PR #228, opened 2018-05-21
Add specific attribute types to basic usage example

thatguysimon:
Hi!
When following the installation steps for the first time, the following warning is shown:
```
WARNING: In previous versions, the Mobility KeyValue backend defaulted to a
text type column, but this behavior is now deprecated and will be removed in
the next release. Either explicitly specify the type by passing type: :text in
each translated model, or set a default option in your configuration.
```
That's because the auto-generated initializer has `default_options[:type]` commented out by default and the basic example of defining attributes doesn't define their types explicitly:
```ruby
class Word < ApplicationRecord
  extend Mobility
  translates :name, :meaning
end
```
This PR simply adds the `type` option to the basic example to avoid the confusing warning for newcomers like myself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,8 @@ of options, like this:
 ```ruby
 class Word < ApplicationRecord
   extend Mobility
-  translates :name, :meaning
+  translates :name,    type: :string
+  translates :meaning, type: :text
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -162,10 +162,21 @@ of options, like this:
 ```ruby
 class Word < ApplicationRecord
   extend Mobility
+  translates :name, :meaning
+end
+```
+
+Note: When using the KeyValue backend, use the options hash to pass each attribute's type:
+
+```ruby
+class Word < ApplicationRecord
+  extend Mobility
   translates :name,    type: :string
   translates :meaning, type: :text
 end
 ```
+
+This is important because this is how Mobility knows to which of the [two translation tables](https://github.com/shioyama/mobility/wiki/KeyValue-Backend) it should save your translation.
 
 You now have translated attributes `name` and `meaning` on the model `Word`.
 You can set their values like you would any other attribute:


shioyama:
Hi @thatguysimon, thanks for the PR! Actually, I did previously have `type` in the examples, but I found that it was confusing since many people who were *not* using the KeyValue backend were including this in their models, which (although harmless) kind of bothered me.

So I removed it, and added it to the initializer generator template as a default which could be enabled. But as you pointed out, this means that a default install hits the error when following the examples, which is also not very good I agree.

How about instead of putting `type` back into the examples, we instead uncomment that line in the initializer generator, wiith a note mentioning that the line can be commented if you are not using the KeyValue backend? This way, people not using this backend may neglect to uncomment the line (which is harmless) but at least their codebase is not littered with unnecessary `type` keys everywhere.

thatguysimon:
So what should the default type be? It's a bit weird to set one default type for all the attributes in the app, don't you think?

Honestly if it was up to me I'd remove the default type option altogether and raise an error if a type wasn't explicitly provided when defining an attribute while using the KeyValue backend. Maybe I'm too harsh but otherwise it just seems like a recipe for bugs and confusion I think.

The readme clearly states in more than one place that the type option is only relevant to the KeyValue backend.

Maybe that basic example in the "Getting started" section should actually be 2 examples? One that uses the type option for the KeyValue backend and another that doesn't (like it is now), for the other backends.

shioyama:
Thanks for the feedback! Let's see...

> So what should the default type be? It's a bit weird to set one default type for all the attributes in the app, don't you think?
> 
> Honestly if it was up to me I'd remove the default type option altogether and raise an error if a type wasn't explicitly provided when defining an attribute while using the KeyValue backend. Maybe I'm too harsh but otherwise it just seems like a recipe for bugs and confusion I think.

Yes, that's basically the change I made in #200 responding to #170, so now it does raise an error if you do not provide a type. But defaults can be set for anything you want (Mobility just merges them into the hash you pass to `translates`). So *preventing* setting a default is not something I want to do, there are valid use cases for that.

To confirm your concern, however, I do agree that it would be better if users were conscious of the impact of a decision of string or text. e.g. the previous default of `text` worried me because this could allocate a *lot* of space for translations when such space is not needed. On the other hand, setting to `string` and then finding out later that you need more space could also be quite problematic...

So although this seems like a little thing, it's potentially actually quite important.

The problem with the suggestion of including two examples, one for KeyValue and one for another backend, is that there are actually multiple places in the readme where a model is defined (I count a total of *eleven*). I stripped the `type` from all of those since as I said I think it's unnecessary for all other backends and detracts from the important points.

To avoid confusion, I added a note about setting the `type` default in the configuration so that KeyValue users can go through the examples and follow exactly what is written, defaulting to whatever type they set in the config.

Do you think adding just one model definition at the beginning, with `type: :string` or something, would help a newcomer? So basically what you changed here, but included *after* the current model definition with a note that for KeyValue you need to specify the type if you haven't set a default?

It's kind of a small point but as you point out, for someone just starting out this little thing could indeed be confusing. I also hesitate because it's important for the user to realize that depending on which type they set, they will be saving the value in a totally different table, and migrating from one to the other later is non-trivial.

thatguysimon:
> Do you think adding just one model definition at the beginning, with type: :string or something, would help a newcomer? So basically what you changed here, but included after the current model definition with a note that for KeyValue you need to specify the type if you haven't set a default?

Exactly! That's what I was suggesting in the last paragraph of my last comment.
Something like:


> ## Getting Started
> Once the install generator has been run to generate translation tables, using Mobility is as easy as adding a few lines to any class you want to translate. Simply pass one or more attribute names to the translates method with a hash of options, like this:
> 
> ```ruby
> class Word < ApplicationRecord
>   extend Mobility
>   translates :name, :meaning
> end
> ```
> 
> Note: When using the KeyValue backend, use the options hash to pass each attribute's type:
> 
> ```ruby
> class Word < ApplicationRecord
>   extend Mobility
>   translates :name     type: :string
>   translates :meaning, type: :text
> end
> ```
> This is important because this is how Mobility knows to which of the [two translation tables](https://github.com/shioyama/mobility/wiki/KeyValue-Backend) it should save your translation.

What do you think?

shioyama:
Looks good! Please update the PR and I'll merge it. Thanks!

thatguysimon:
Done, have a look.
Thanks!

shioyama:
Merged in f7d095d80da184e9f8dbb570ca039836668ecb2a, thanks! :smile:

PR #229, opened 2018-05-22
Remove Mobility::Interface

shioyama:
Mobility currently has an interface for accessing certain model-global methods like `names` which returns the attribute names on the model (for all backends). This is accessible via `Post.mobility`.

However, with the extensive changes in #225, I'm finding that having both the class query scope `Post.i18n` *and* the interface on `Post.mobility` is getting pretty clumsy. Here I am going to try to move this interface stuff into the `Attributes` module builder so that class methods are defined on the model as soon as any attributes are defined, without using a memoized instance variable (which requires extra precautions when `dup`ing etc.)

The goal here is to have this type of interface for accessors and backends:

```ruby
  extend Mobility
  translates :title
end

post = Post.new
post.mobility_backends
#=> {}
post.title = "foo"
post.mobility_backends
#=> {:title=> <memoized instance of backend class>}
Post.mobility_backend_class(:title)
#=> #<Mobility::Backends::ActiveRecord::...>
```

In the `i18n` scope, you also get the `[]` shortcut to access nodes:

```ruby
Post.i18n[:title]
#=> Arel node for `title` attribute
```

Although I made querying into a plugin (on the `master` branch only), I'm going to convert it into a normal module which is included by default when you extend `Mobility`. Querying in the end is something global that should only enabled on an entire model, or not at all.


shioyama:
I've gone with `backend_node` as the name of the method to get the node on the query scope:

```ruby
Post.i18n.backend_node[:title]
#=> Arel node for title attribute
```

Anything generic like `[]` will possibly conflict with other scope methods added by gems, etc.

PR #233, opened 2018-05-27
Allow passing locale to query methods

shioyama:
Replaces #210, much simpler now.

### DIFF
--- a/lib/mobility/plugins/active_record/query.rb
+++ b/lib/mobility/plugins/active_record/query.rb
@@ -113,7 +113,7 @@ enabled for any one attribute on the model.
               def build(scope, where_opts, invert: false)
                 return yield unless Hash === where_opts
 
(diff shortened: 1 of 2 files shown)


shioyama:
Here's what this does:

```ruby
Post.i18n.where(title: "foo", locale: :ja)
#=> returns post with Japanese title "foo"
```

The interface here may be slightly confusing, since you pass the locale as an option to `i18n` when setting the locale for block queries:

```ruby
Post.i18n(locale: :ja) { title.eq("foo") }
#=> returns post with Japanese title "foo"
```

but this locale keyword to `i18n` will be ignored when used with query methods:

```ruby
Post.i18n(locale:ja).where(title: "foo")
#=> ignores locale and returns post with title "foo" in current locale
```

For the above case to work, we would have to store the locale in the chained relation, which I don't think I really want to do.

PR #236, opened 2018-05-29
Implement case-insensitive uniqueness validation

shioyama:
Replaces #221


shioyama:
Pleasantly surprised to find that this was very easy to implement with Arel support.

PR #240, opened 2018-05-30
Support equality predicates between jsonb nodes

shioyama:
Previously it was not possible to create equality predicates between jsonb nodes. This fixes the issue, so you can do:

```ruby
Post.i18n { title.eq(content) }
```

... and Mobility will return all posts whose `title` and `content` are the same, for Jsonb (and Container) backends (already worked for other backends). I think this is the last issue to fix before releasing 0.7.0 :tada:

### DIFF
--- a/lib/mobility/arel/nodes/pg_ops.rb
+++ b/lib/mobility/arel/nodes/pg_ops.rb
@@ -30,9 +30,13 @@ module Mobility
       end
 
       # Needed for AR 4.2, can be removed when support is deprecated
-      HstoreDashArrow.class_eval do
-        def quoted_node other
-          other && super
+      if ::ActiveRecord::VERSION::STRING < '5.0'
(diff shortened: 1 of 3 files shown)


lib/mobility/arel/nodes/pg_ops.rb
```diff
@@ -29,54 +29,66 @@ def lower
...
+        def to_question
+          JsonbQuestion.new(left, right)
+        end
+
+        def eq other
+          case other
+          when NilClass
+            to_question.not
+          when Integer, Array, Hash
+            to_dash_arrow.eq(other.to_json)
```
shioyama: Previously Jsonb used the default `->` operator, but this is problematic in general, so instead we now use `->>` which returns a string and does not require casting. In case we are calling `eq` with an argument that is an integer, array or hash, we convert the argument to json and compare using `->` as before.

Non-string translations are an edge case, but one that I want to support.

Issue #243, opened 2018-05-30
0.7.0 breaks translates without options


pwim:
``` ruby
class Question < ApplicationRecord
  translates :text
end
```
results in a syntax error under 0.7.0:
```
  /usr/local/var/rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/mobility-0.7.0/lib/mobility/plugins/locale_accessors.rb:60: syntax error, unexpected '{', expecting ';' or '\n'
          def name_{:en=>:ja, :ja=>:en}(options =...
                   ^
```
Switching to having an empty options works around the bug.
``` ruby
class Question < ApplicationRecord
  translates :text, {}
end
```
I noticed all the test cases have an explicit options hash, but I'm assuming this isn't required.

glebm:
Same with 0.7.2

shioyama:
Sorry, somehow missed this issue. I'll fix it today.

shioyama:
Hmm strange, from the error above `I18n.available_locales` seems to be returning the fallbacks, but I don't see this on a dummy app. I'll have to dig further.

shioyama:
Ok I created a clean dummy app with a single model `Question`, with:

```ruby
class ApplicationRecord < ActiveRecord::Base
  self.abstract_class = true
  extend Mobility
end
```

```ruby
class Question < ApplicationRecord
  translates :text
end
```

And I can't reproduce this... model loads fine and I can instantiate and create a question with a `text` attribute value. Must be something in your config that is triggering it.

glebm:
It looks like this is caused by Mobility-enabled readers getting defined before `I18n` is initialized by Rails:

```console
> I18n.available_locales
[:en, :ru, :ro, :"zh-TW", :id, :tr, :hu, :"zh-CN", :it, :nl, :de, :da, :es, :ja, :fr, :cs, :"pt-BR", :sk, :"en-GB", :"fi-FI", :"en-US", :pl, :"en-PAK", :ca, :bg, :ko, :"en-NG", :uk, :"nb-NO", :"en-SG", :"ca-CAT", :"en-ZA", :"de-AT", :"en-UG", :"en-IND", :pt, :"da-DK", :vi, :"en-BORK", :"en-CA", :"en-au-ocker", :he, :"en-AU", :sv, :fa, :"es-MX", :nep, :"de-CH", :"en-MS", :"en-NZ"]

> Rails.application.config.i18n.available_locales
["en", "es"]
```

[...]

PR #253, opened 2018-06-07
Normalize locales with multiple dashes and raise exception if locale accessor cannot be defined

shioyama:
Fixes issue mentioned [here](https://github.com/shioyama/mobility/issues/243#issuecomment-395302188).

### DIFF
--- a/lib/mobility.rb
+++ b/lib/mobility.rb
@@ -37,6 +37,8 @@ module Mobility
 
   # General error for version compatibility conflicts
   class VersionNotSupportedError < ArgumentError; end
+  CALL_COMPILABLE_REGEXP = /\A[a-zA-Z_]\w*[!?]?\z/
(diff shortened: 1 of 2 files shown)


shioyama:
@glebm I'm going to merge and release this change, which will allow using Faker locales in tests without issue. Doesn't solve the existing problem of Mobility not picking up Rails i18n available locales, I'll deal with that separately.

shioyama:
We could handle more complex method names using `define_method`, if that's something anybody really wants (that's what AttributeMethods [does](https://github.com/shioyama/rails/blob/847342c25c61acaea988430dc3ab66a82e3ed486/activemodel/lib/active_model/attribute_methods.rb#L366-L370), but I'm assuming that other than weird locales, we can assume attributes themselves only include word characters (i.e. a-zA-Z0-9_).

PR #254, opened 2018-06-08
Prefer Rails available_locales if loaded

shioyama:
For background, see: https://github.com/shioyama/mobility/issues/243#issuecomment-395302188

I don't like having Rails-specific code sprinkled around, but this seems fairly minimal and probably will avoid weird issues in Rails applications.

### DIFF
--- a/lib/mobility.rb
+++ b/lib/mobility.rb
@@ -117,7 +117,7 @@ module Mobility
     # Sets Mobility locale
     # @param [Symbol] locale Locale to set
     # @raise [InvalidLocale] if locale is nil or not in
-    #   +I18n.available_locales+ (if +I18n.enforce_available_locales+ is +true+)
+    #   +Mobility.available_locales+ (if +I18n.enforce_available_locales+ is +true+)
     # @return [Symbol] Locale
     def locale=(locale)
       set_locale(locale)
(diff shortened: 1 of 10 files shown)


shioyama:
@glebm This PR will use Rails-defined available locales if they are loaded when required. I'm a bit hesitant about this because I'm wondering if now `I18n.locale_available?(locale)` will check against the wrong locales list... but I don't really want to define yet another method on Mobility.

In any case, I think it's important that Rails configuration is properly picked up by Mobility so this seems sensible. I think since it's not really urgent I'll just merge once specs pass and release in the next minor.

glebm:
> I'm wondering if now `I18n.locale_available?(locale)` will check against the wrong locales list.

This would happen if `Mobility.set_locale` or a locale accessor instance method is called in an initializer. This is unlikely.

It might be nice to do this anyway because these issues are hard to diagnose.

```ruby
def locale_available?
  (@available_locales_set ||= Set.new(available_locales.map(&:to_s)).include?(locale.to_s)
end
```

shioyama:
```ruby
(@available_locales_set ||= Set.new(available_locales.map(&:to_s)).include?(locale.to_s)
```

I prefer not to memoize booleans :wink: 

But also, I think handling this rare case could cause more complications, so I think I'm going to leave it with just the change to `available_locales`. I feel that really, this is a Rails issue, not a Mobility issue. Rails is using I18n, so it should use `I18n.available_locales` rather than its own (separate) configuration, or at least it make them identical.

Anyway thanks for discovering this, although it probably doesn't cause problems to most people it means that very likely many applications were defining *many* locale accessors that would never ever be used, which is wasteful.

glebm:
That's fair, though rails is unlikely to change this at this point as it would cause backwards incompatibility.

The code memoizes the set not the Boolean though!

shioyama:
> The code memoizes the set not the Boolean though!

Ah, right you are. Too many parens for my mind to process...

PR #255, opened 2018-06-08
Refactor Sequel querying into plugin

shioyama:
Still a long way to go...


shioyama:
Json/Jsonb/Hstore/Container and Column backends all working now. Hardest two are left: KeyValue and Table.

shioyama:
All backends are now passing for me. Just need to clean this up and it should be almost ready to merge, so we can release 0.8.

PR #261, opened 2018-08-27
Support order clause on translated queries

### LINKED ISSUE #171: Support order clause on translated queries
Mobility does not currently support an `order` clause on translated queries. We should see how much work it would be to support it in both ActiveRecord and Sequel.

## Context

You can currently query on a translated attribute with the `i18n` scope, like this:

```ruby
Post.i18n.where(title: "foo")
```

and Mobility will figure out what query is actually needed to fetch a post with  a title "foo" in the current locale.

However, ordering results is not supported:

```ruby
Post.i18n.where(title: "foo").order(:title)
# raises: ActiveRecord::StatementInvalid: PG::UndefinedColumn: ERROR: column posts.title does not exist
```

This should be possible to implement without too much work.

Ref: #169

shioyama:
Fixes #171

### DIFF
--- a/lib/mobility/attributes.rb
+++ b/lib/mobility/attributes.rb
@@ -281,6 +281,13 @@ EOL
         mobility_modules.map(&:names).flatten.uniq
       end
 
+      # Return true if attribute name is translated on this model.
+      # @param [String, Symbol] Attribute name
+      # @return [Boolean]
+      def mobility_attribute?(name)
+        mobility_attributes.include?(name.to_s)
+      end
+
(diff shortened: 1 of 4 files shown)


shioyama:
@eiapopeia This is very much a WIP, but have a look and give it a try if you're interested.

shioyama:
This is almost ready to go, I'll just squash commits and clean up specs a bit later today or tomorrow and merge. But ideally would like to get some people to check it works for them before releasing.

shioyama:
Merged!

PR #265, opened 2018-08-29
Yield to block (super) if query map is the identity function

shioyama:
This kind of subtle, but basically if the query map here is the identity function, it means we did not match *any* keys to module keys, so there is nothing to do here. Checking if it is nil is meaningless since inject in this case will always return something.


shioyama:
As an aside, I wonder why Ruby does not provide an identity map somewhere to use. It feels weird to have to define `IDENTITY = ->(x) { x }.freeze`...

PR #269, opened 2018-09-11
Support block query format with Sequel

shioyama:
This brings Sequel support almost up to the same level as ActiveRecord (except missing support for `order` in the Query plugin, which I may try to add before releasing 0.8.0).


shioyama:
Github is not showing Travis builds for some reason...

PR #270, opened 2018-09-12
Only use unique plugin names

shioyama:
I debated using a set here, but this should basically have the same effect and doesn't require `require`-ing `set`.


shioyama:
Closing this, don't really think it's necessary.

Issue #226, opened 2018-05-17
Using Mobility with JSON backend and MySQL 5.7.22 in production


enthusiasmus:
Although the documentation doesn't mention one can use the json backend with mysql, we are using it already in production. It works as if we were using postgres. Maybe you can add MySQL with JSON backend to your test suite and update the documentation?

Mobility version: 0.6.0
Mobility backend: `:json`

Thanks,
best regards
Lukas

shioyama:
Wow, that's really interesting! Don't often get issues for features that you didn't know are already done. Out of curiosity, do you use querying on the json columns as well?

shioyama:
Wow, so it seems that [indeed](https://dev.mysql.com/doc/refman/5.7/en/json-search-functions.html#operator_json-inline-path), the syntax for querying mysql columns is almost identical to pg (or at least, MySQL has aliases that make it behave the same way). As of 5.7.13, you can use `column ->> path`, which is what Mobility uses for the Json backend. So I could imagine everything working out-of-the-box, which is really nice.

Ideally I'd like to add tests for MySQL before adding it to the docs, not sure Travis will work and locally I use MariaDB which I believe is not compatible with MySQL JSON, or at least works differently... in any case, could you confirm whether querying does indeed work for you? That would make a difference in whether to list it on the wiki/readme or not.

enthusiasmus:
Yep, searching works like expected. Although we currently don't query entities by translated attributes in production. Here is an example against our database:

```mysql> SELECT label, label->"$.de" FROM spree_site_links;
+------------------------------------+---------------+
| label                              | label->"$.de" |
+------------------------------------+---------------+
| {"de": "Kontakt", "fr": "contact"} | "Kontakt"     |
| {"de": "Presse"}                   | "Presse"      |
| {"de": "Impressum"}                | "Impressum"   |
| {"de": "AGB"}                      | "AGB"         |
| {"de": "Anbieter"}                 | "Anbieter"    |
| {"de": "Events"}                   | "Events"      |
+------------------------------------+---------------+
6 rows in set (0.00 sec)

mysql> SELECT label, label->"$.de" FROM spree_site_links WHERE JSON_EXTRACT(label, "$.de") = "Kontakt";
+------------------------------------+---------------+
| label                              | label->"$.de" |
+------------------------------------+---------------+
| {"de": "Kontakt", "fr": "contact"} | "Kontakt"     |
+------------------------------------+---------------+
1 row in set (0.00 sec)```

[...]

PR #271, opened 2018-09-14
Add mysql json support (wip)

kule:
This is an initial stab at getting mysql json tests passing (related https://github.com/shioyama/mobility/issues/226).

This isn't ready for use yet and is more to get discussion going - this will break postgres specs as I've commented out pg_opts for now.

Things that needed changing to get the mysql tests passing:

- Rather then 'en', json keys must be in the format '$.en'.
- Because dashes aren't valid ECMAScript identifiers these must be escaped so '$.en-GB' needs to become '$."en-GB"'
- You can't have a default value on a json column
- The first set of like/ilike specs (with case-insensitivity) fail for mysql because the json appears to be case-sensitive regardless of collation.  Would have to use lower() by the looks of things...

TODO - would appreciate some thoughts or ideas on these:

- Uncomment pg_opts and refactor so mysql_opts can be loaded too
- Refactor quoting - this'll probably need escaping properly
- Decide whether to ignore or fix the json case-insensitivity tests

### DIFF
--- a/lib/mobility/arel/nodes/mysql_ops.rb
+++ b/lib/mobility/arel/nodes/mysql_ops.rb
@@ -1,0 +1,50 @@
+# frozen-string-literal: true
+require "mobility/arel"
+
+module Mobility
+  module Arel
+    module Nodes
+      %w[
+        JsonDashArrow
+        JsonDashDoubleArrow
+      ].each do |name|
+        const_set name, (Class.new(Binary) do
+          include ::Arel::Expressions
+          include ::Arel::Predications
+          include ::Arel::OrderPredications
+          include ::Arel::AliasPredication
+
+          def lower
+            super self
+          end
+        end)
+      end
+
+      class Json < JsonDashDoubleArrow; end
+
+      class JsonContainer < Json
+        def initialize column, locale, attr
+          super(Arel::Nodes::JsonDashArrow.new(column, locale), attr)
+        end
+      end
+    end
+
+    module Visitors
+      def visit_Mobility_Arel_Nodes_JsonDashArrow o, a
+        json_infix o, a, '->'
+      end
+
+      def visit_Mobility_Arel_Nodes_JsonDashDoubleArrow o, a
+        json_infix o, a, '->>'
+      end
+
+      private
+
+      def json_infix o, a, opr
+        visit(Nodes::Grouping.new(::Arel::Nodes::InfixOperation.new(opr, o.left, o.right)), a)
+      end
+    end
+
+    ::Arel::Visitors::MySQL.include Visitors
+  end
+end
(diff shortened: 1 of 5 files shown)


shioyama:
Thanks very much! Really happy to see this, looks great. I'll have a look in the next few days and get back about where to go from here.

shioyama:
> Looks like things don’t play so nice in rails 4 either so I’ll have a look see what’s needed.

Don't worry about that, it's absolutely fine not to support Rails 4 on this feature.

You can use the metadata `rails_version_geq: 5.0` to mark specs for mysql json that fail in Rails 4, like this:

https://github.com/shioyama/mobility/blob/1686f5d933da1f32dceed896cfc76f1f0b6b895c/spec/support/shared_examples/serialization_examples.rb#L57-L65

shioyama:
> Because dashes aren't valid ECMAScript identifiers these must be escaped so '$.en-GB' needs to become '$."en-GB"'

I believe underscore is allowed as an identifier, so couldn't we use `'$.en_gb'` instead? i.e. `Mobility.normalize_locale(locale)`.

> You can't have a default value on a json column

Hmm. With postgres you can set the deafault to an empty hash, is this not possible with MySQL? It makes things slightly simpler if you can do this.

> The first set of like/ilike specs (with case-insensitivity) fail for mysql because the json appears to be case-sensitive regardless of collation (EDIT: so currently I've skipped these for mysql). Would have to use lower() by the looks of things...

That's totally fine!

shioyama:
> Uncomment pg_opts and refactor so mysql_opts can be loaded too

I think we should probably namespace the pg-specific constants, maybe we can just prepend `Pg` to their names?

i.e. instead of `Mobility::Arel::Nodes::Json`, we call that `Mobility::Arel::Nodes::PgJson`, etc? Then you could also define `Mobility::Arel::Nodes::MysqlJson` maybe? Not sure about the naming.

lib/mobility/arel/nodes/mysql_ops.rb
```diff
@@ -0,0 +1,50 @@
...
+        end)
+      end
+
+      class Json < JsonDashDoubleArrow; end
+
+      class JsonContainer < Json
+        def initialize column, locale, attr
+          super(Arel::Nodes::JsonDashArrow.new(column, locale), attr)
+        end
+      end
```
shioyama: It's great that you're trying to support a container format for MySQL Json, but I think it will make things a bit complicated for this first PR. How about removing this and leaving it for another PR later?

shioyama:
@kule I made some quick comments, let me know what you think. I'll try to have another look later this week.

kule:
@shioyama thanks I'll see if I can get time towards the end of the week to make these changes.

RE default values - no it's a limitation with MySQL unfortunately: https://dev.mysql.com/doc/refman/8.0/en/json.html (about 4th paragraph down).  MySQL has the same issue with TEXT & BLOB fields, I think it's due to the way it stores them.

PR #276, opened 2018-09-17
Mention required ruby version change in 0.8.0 changelog

valscion:
We're still running on a bit outdated Ruby version, so the required ruby version change in https://github.com/shioyama/mobility/pull/242 was an unfortunate surprise to us given that it was not mentioned in the changelog.

So here's an update to the changelog so others won't get bitten by the same change :smile:

Looks like we should upgrade Ruby quite soon 😅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
   [#268](https://github.com/shioyama/mobility/pull/267))
 * Default locale to Mobility.locale in apply_scope
   ([#263](https://github.com/shioyama/mobility/pull/263))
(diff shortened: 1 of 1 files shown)


shioyama:
Thanks very much! You're right this should have been mentioned in the changelog. 2.2 is not supported anymore which is why I dropped support for it, but in practice Mobility will probably work fine with it (if you change the gemspec in a fork or something).

Ref: #242.

valscion:
Yeah we're at fault here for running an unmaintained version ourselves. Thanks for the quick merge and for this gem :)

PR #280, opened 2018-09-20
Bump i18n dependency

### LINKED ISSUE #278: Compatibility with i18n v1.1
Adding Mobility to my projects Gemfile forced a downgrade on the gem *i18n* from *v1.1.0* to *v1.0.1*. Is this really necessary? Maybe the gemspec is just too strict?

```
Bundler could not find compatible versions for gem "i18n":
  In snapshot (Gemfile.lock):                                                                                                                                
    i18n (= 1.1.0)                                                                                                                                           

  In Gemfile:                                                                                                                                                
    mobility (~> 0.8.0) was resolved to 0.8.0, which depends on                                                                                              
      i18n (< 1.1, >= 0.6.10)
``` 

If i18n uses semantic versioning, `i18n (< 2, >= 0.6.10)` should work in the gemspec. In case it's really not working, I hope you support *i18n 1.1* from the next version on.

shioyama:
Fixes #278

### DIFF
--- a/spec/integration/active_record_compatibility_spec.rb
+++ b/spec/integration/active_record_compatibility_spec.rb
@@ -77,11 +77,6 @@ describe "ActiveRecord compatibility", orm: :active_record do
   describe "fallbacks" do
     let!(:post) { FallbackPost.create(title: "foo title") }
 
-    it "falls through to default locale with no options" do
-      Mobility.locale = :ja
-      expect(post.title).to eq("foo title")
-    end
-
     it "does not fall through to default locale when fallback: false option passed in" do
       Mobility.locale = :ja
       expect(post.title(fallback: false)).to eq(nil)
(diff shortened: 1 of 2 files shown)


shioyama:
@omitter so it does seem there is a small regression with I18n 1.1, two specs related to fallbacks are failing. I'll have to look into it, but if you want to have a look that would help speed things up.

shioyama:
See here: https://travis-ci.org/shioyama/mobility/jobs/431276992

omitter:
Checking the linked commit, it looks like they simply deleted the fall back tests ([here](https://github.com/svenfuchs/i18n/pull/415/commits/8ddfb17d73f2fd7078c67ce9da5558fdd2b96c42#diff-a2055b8f77e6d8fe48b7d70447bc5739)). But I guess it's not a good idea to do here too.

PR #282, opened 2018-09-21
migration generator compatibility with rails>=5.1

omitter:
Also added foreign key


shioyama:
Thanks for this! It looks good to me. Can you update the spec to pass? You will need to update [this line](https://github.com/shioyama/mobility/blob/99c85ef6141903c6e188f249eb15bc8bb1806533/spec/generators/rails/mobility/translations_generator_spec.rb#L71).

omitter:
Ok, looks fine now. I think the one crash is due to the build system (timeout).

PR #285, opened 2018-09-26
Implement select/group in AR query plugin

### LINKED ISSUE #275: Column backend does not support pluck or select
## Context

I'm using `column` backend strategy in mobility and I can perform queries based on my translated column using `find_by` and even `where` but what intrigues me is that `select` or `pluck` does not know how to handle my column name.

I'm not considering this a problem but in fact a feature not implemented, if there is an already approach to handle this I'm sorry for the issue but I really didn't find nothing about in docs and stackoverflow.

## Expected Behavior

I expect mobility to handle my column name when I use `select` or `pluck` methods.

```ruby
Mobility.locale = :jp
Post.all.pluck(:title) # select title_jp from posts
```

## Actual Behavior

```ruby
Post.all.pluck(:title) # ERROR:  column "title" does not exist
```

shioyama:
Fixes #275

### DIFF
--- a/lib/mobility/arel.rb
+++ b/lib/mobility/arel.rb
@@ -4,7 +4,36 @@ require "mobility/arel/visitor"
 
 module Mobility
   module Arel
+    module MobilityExpressions
+      include ::Arel::Expressions
+
+      # @note This is necessary in order to ensure that when a translated
+      #   attribute is selected with an alias using +AS+, the resulting
+      #   expression can still be counted without blowing up.
+      #
+      #   Extending +::Arel::Expressions+ is necessary to convince ActiveRecord
+      #   that this node should not be stringified, which otherwise would
+      #   result in garbage SQL.
+      #
+      # @see https://github.com/rails/rails/blob/847342c25c61acaea988430dc3ab66a82e3ed486/activerecord/lib/active_record/relation/calculations.rb#L261
+      def as(*)
+        super
+          .extend(::Arel::Expressions)
+          .extend(Countable)
+      end
+
+      module Countable
+        # @note This allows expressions with selected translated attributes to
+        #   be counted.
+        def count(*args)
+          left.count(*args)
+        end
+      end
+    end
+
     class Attribute < ::Arel::Attributes::Attribute
+      include MobilityExpressions
+
       attr_reader :backend_class
       attr_reader :locale
       attr_reader :attribute_name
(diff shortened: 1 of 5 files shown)


shioyama:
Note: `select` is quite tricky, because in ActiveRecord attribute methods on the results of a select query are handled by `method_misisng` in a [sparsely-documented feature](https://github.com/shioyama/rails/blob/847342c25c61acaea988430dc3ab66a82e3ed486/activemodel/lib/active_model/attribute_methods.rb#L415-L424) dating back all the way to the early days of Rails.

The tricky part is that Mobility has already defined these accessor methods, so it can't easily delegate to method missing like AR does internally in this case. I'm not yet sure how to handle this.

Most stuff works simply by using the "as <attribute name>" format, where `attribute name` here is the name of the translated attribute. But the accessors (i.e. `post.title`) do not, for the reason above.

lib/mobility/arel.rb
```diff
@@ -4,7 +4,25 @@
 
 module Mobility
   module Arel
+    module MobilityExpressions
+      include ::Arel::Expressions
+
+      def as(*)
+        super
+          .extend(::Arel::Expressions)
```
shioyama: This is necessary due to [this line](https://github.com/shioyama/rails/blob/847342c25c61acaea988430dc3ab66a82e3ed486/activerecord/lib/active_record/relation/calculations.rb#L261). We might not need it though.

lib/mobility/plugins/active_record/query.rb
```diff
@@ -22,6 +22,28 @@ def apply(attributes)
...
+        #   Mobility attribute from the attribute accessor. We use the
+        #   "SELECT ... AS ..." form in the query, so ActiveRecord assigns
+        #   the returned value to the attributes hash, but in order to
+        #   actually return it we need to intercept the read here using
+        #   +instance_variable_get+. There should be a cleaner/safer way to
+        #   do this...
+        # TODO: Improve this.
+        def read(locale, **)
+          attr_set = model.instance_variable_get(:@attributes)
+          if attr_set && attr_set.key?(attr_alias = Query.attribute_alias(attribute))
```
shioyama: I've switched to using an alias when selecting, so `SELECT ... AS __mobility_title__`. This avoids conflicts with other stuff in ActiveRecord, but still passes grouping specs.

I still don't like the `instance_variable_get` here...

lib/mobility/arel.rb
shioyama: Seems we do need it.

PR #288, opened 2018-09-30
Handle untranslated scopes in AR query module

### LINKED ISSUE #266: joins does not works with multiple levels
First of all, thank you for the great gem.

I found a case that joins does not works. When attributes is on the root of hash, it works, but with join we need a second level attributes, so the joins is not triggered.

## Context

I'm using backend key/value with `query_order` branch on MySQL.

## Behaviors

This code works with `slug` on root of the hash condition:

```
Category.i18n.joins(:category_profiles).where(slug: "the-title").to_sql
```

```
"SELECT `categories`.* FROM `categories` INNER JOIN `categories_profiles` ON `categories_profiles`.`category_id` = `categories`.`id` INNER JOIN `mobility_string_translations` `Category_slug_pt_br_string_translations` ON `Category_slug_pt_br_string_translations`.`key` = 'slug' AND `Category_slug_pt_br_string_translations`.`locale` = 'pt-BR' AND `Category_slug_pt_br_string_translations`.`translatable_type` = 'Category' AND `Category_slug_pt_br_string_translations`.`translatable_id` = `categories`.`id` WHERE `Category_slug_pt_br_string_translations`.`value` = 'the-title'"
```

But when I scope the `slug` into the hash, it does not works:

```
Category.i18n.joins(:category_profiles).where(categories: { slug: "the-title" }).to_sql
```

```
"SELECT `categories`.* FROM `categories` INNER JOIN `categories_profiles` ON `categories_profiles`.`category_id` = `categories`.`id` WHERE `categories`.`slug` = 'the-title'"
```

## Possible Fix

Maybe always trigger joins when `i18n` is used can solve the problem, mas if people to use it as `default_scope` will degrade performance.

I tried to use `join_mobility_model_translations` and `join_translations` but this methods are missing.

shioyama:
Fixes #266


shioyama:
Something wrong with the specs related to `stub_const`, but pretty sure the fix is fine. I'll investigate later.

Issue #258, opened 2018-06-26
NameError (uninitialized constant Mobility::ActiveRecord::VERSION)


vollnhals:
I did a plain install into my existing rails 5.2 (activerecord) project

```
gem "mobility", "~> 0.7.5"
```

and added Mobility to my `Event` class.

```
class Event < ApplicationRecord
  extend Mobility

  # relevant for discussion
  multi_tenant :tenant
end
```

## Expected Behavior
My app still works.

## Actual Behavior
`Event.new` does not work anymore.

```
NameError (uninitialized constant Mobility::ActiveRecord::VERSION)
```

This happens I think because `extend Mobility` does `include Mobility::ActiveRecord`.

When the call to `multi_tenant` from the `activerecord-multi-tenant` gem is called, it wants to look up `ActiveRecord` but it will find `Mobility::ActiveRecord` now.

So by including `mobility` into my project, the project broke.

## Possible Fix
The same issue was discussed in #69 . But as far a I can see the issue was closed prematurely because the cause is not `extend Mobility` , it is the `include Mobility::ActiveRecord`.

Also see https://github.com/citusdata/activerecord-multi-tenant/pull/42 for a work around .

shioyama:
Thanks, very good bug report! You're right that this was previously encountered in #69. I wonder if the only way to fix it is to rename `Mobility::ActiveRecord` to something else. I believe this is why e.g. [Roda](https://github.com/jeremyevans/roda) uses `Roda::RodaRequest` rather than `Roda::Request`.

shioyama:
I suppose the solution would be for all class names which potentially conflict with base classes (`Mobility::ActiveRecord`, `Mobility::ActiveModel`, `Mobility::Arel` and `Mobility::Sequel`), we could append the string "Mobility" to the class, so we'd have:

- `Mobility::MobilityActiveRecord`
- `Mobility::MobilityActiveModel`
- `Mobility::MobilityArel`
- `Mobility::MobilitySequel`

I've avoided doing this because it feels like it shouldn't be necessary, and I do think that it's the gem owner's responsibility to always call the base class explicitly (Mobility does this everywhere). But OTOH this could cause very strange bugs which might be harder to track down than the one in this issue.

If anyone has any thoughts or ideas, I'd be eager to hear before making this change.

cc @pwim

pwim:
[...]

PR #291, opened 2018-10-12
[WIP] Add regression spec for issue #258

shioyama:
Test for #258. Demonstrates the issue.

It seems that the names for any constants under `Mobility` need to be chosen carefully to not potentially conflict with other gems, due to the fact that we `extend Mobility`.

I need to think about this a bit more, since it would mean a shake up in a lot of constants (although not the backends or plugins themselves I believe, since they are under `Mobility::Backends`, and `Mobility::Plugins`, unless we rename those constants too...).

If anyone has any experience with this kind of lookup issue I'd be keen to get suggestions on this.

### DIFF
--- a/spec/mobility/active_record_spec.rb
+++ b/spec/mobility/active_record_spec.rb
@@ -15,4 +15,14 @@ describe Mobility::ActiveRecord, orm: :active_record do
       expect(Article.new.translated_attribute_names).to eq(["foo"])
     end
   end
+
+  # regression spec for https://github.com/shioyama/mobility/issues/258
+  describe "name error" do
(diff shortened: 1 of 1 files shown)


shioyama:
Maybe we could support an alternative usage which would instead include an anonymous module that would do the same thing as `extend Mobility`. If it was anonymous I don't believe this problem would be triggered.

This is maybe why a gem like [Virtus](https://github.com/solnic/virtus) uses the `include Virtus.model` pattern (where `model` returns `Module.new` with some stuff added).

shioyama:
Fixed by #464

PR #307, opened 2018-11-23
Fix destroying empty translations in AR table backend

raulr:
Destroying records during `each` messes up the loop internal pointer, so when there is more than one empty translation some of them are not destroyed.


shioyama:
Thanks very much! I guess the spec (as you've updated it) fails without the code change, right? If so looks good to me.

raulr:
Thats right, I added another empty translation in the specs so it would fail without the fix.

shioyama:
Perfect, thanks!

shioyama:
Released in 0.8.6.

PR #322, opened 2019-04-21
Test against rails master

shioyama:


shioyama:
Pretty sure failures are from rails/rails#25933. This may be very hard to get around.

shioyama:
Pretty sure failures are from rails/rails#35933. This may be very hard to get around.

PR #336, opened 2019-08-28
CI: Use newer 2.5.6, 2.6.4

olleolleolle:
This PR updates the CI matrix by using the latest version of Ruby.

(2.4.7 had install problems today on `rvm` in Travis, so I left that as a next update.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: ruby
 cache: bundler
 rvm:
   - 2.4.5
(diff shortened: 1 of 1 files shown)


shioyama:
Thanks!

doits:
FYI: I think you can simply omit the last patch number and write `2.5` and `2.6` etc. to always use the *latest available* ruby, so you automatically stay up to date with latest ruby.

PR #343, opened 2019-10-17
Fix Dirty plugin with Rails 6

### LINKED ISSUE #323: Dirty tracking does not work with ActiveRecord > 6.0.0.beta.3

shioyama:
Fixes #323

See my [comment below](https://github.com/shioyama/mobility/pull/343#issuecomment-544369012) for a summary of changes.

### DIFF
--- a/spec/mobility/plugins/active_model/dirty_spec.rb
+++ b/spec/mobility/plugins/active_model/dirty_spec.rb
@@ -142,7 +142,7 @@ describe "Mobility::Plugins::ActiveModel::Dirty", orm: :active_record do
   end
 
   describe "suffix methods" do
-    it "defines suffix methods on translated attribute", rails_version_geq: '5.0' do
+    it "defines suffix methods on translated attribute" do
       instance.title = "foo"
       instance.save
 
@@ -154,12 +154,16 @@ describe "Mobility::Plugins::ActiveModel::Dirty", orm: :active_record do
         expect(instance.title_was).to eq("foo")
 
         instance.save
-        if ENV['RAILS_VERSION'].present? && ENV['RAILS_VERSION'] < '5.0'
-          expect(instance.title_changed?).to eq(nil)
-        else
+        expect(instance.title_changed?).to eq(false)
+
+        if ENV['RAILS_VERSION'].present? && ENV['RAILS_VERSION'] >= '5.0'
           expect(instance.title_previously_changed?).to eq(true)
           expect(instance.title_previous_change).to eq(["foo", "bar"])
           expect(instance.title_changed?).to eq(false)
+
+          if ENV['RAILS_VERSION'].present? && ENV['RAILS_VERSION'] >= '6.0'
+            expect(instance.title_previously_was).to eq('foo')
+          end
         end
 
         instance.title_will_change!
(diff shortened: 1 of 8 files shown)


shioyama:
This will not fix dirty issues for 99% of Mobility users, since of course most of you are using it with Active **Record**, and this just fixes the issue for Active **Model**. But it's a step. And yet another ugly versioned patch...

shioyama:
Ok, I got something which works in Rails/AR 6. I had to define a custom mutation tracker, and completely change the way Mobility handles tracking of dirty changes, but this works and should be (I think) a bit more robust to changes in ActiveRecord/ActiveModel.

I'll try to clean it up a bit, and then add back necessary hacks to get tests passing in other versions of ActiveRecord. For now I'm only watching Rails 6 tests, other stuff can be fixed once those are passing.

shioyama:
This fixes the problems with dirty in Rails 6. It adds a lot of lines of code, mostly because I entirely re-implemented the plugin for Rails 6, but it works and I think it will be more robust to changes.

Once everything is passing I'll clean up commits and release this as 0.8.9.

shioyama:
Tests passing! I'm going to improve this a bit and should be able to merge/release on Monday.

shioyama:
This is done and specs are passing. It adds about 150 lines of code (including some added specs), but should be more robust to changes in Rails.

## Problem

In the past, Mobility has attempted to leverage ivar-backed dirty tracking in ActiveModel, but this was always tricky because it's never been seriously supported in ActiveRecord. This is a longtime gripe of mine (that AM and AR are so heavily coupled), but I think it's better to just accept the situation and find an optimal solution. This became more pressing with rails/rails#35933, which optimized dirty tracking methods by simply doing away with ivar-backed tracking altogether.

## Solution

Instead of leveraging code in ActiveModel and ActiveRecord, which makes the plugin fragile to changes there, I just created a new "mutation tracker" similar to the ones defined in ActiveModel. I coded this from scratch rather than try to inherit since inheritance also makes the code fragile.

The mutation tracker is called from overrides to all the dirty methods to check if a translated attribute has changed, etc. Since we now explicitly define overrides, and use an entirely separate mutation tracker, changes in Rails itself should have a much narrower impact on Mobility. An indication of this is that the new code has much less version-conditionals than the previous code did (most conditionals are just used to say which methods should be defined for which Rails versions, since dirty patterns have been added several times).

I'll merge this and release it as 0.8.9. I hesitated about version numbering, but since this is not really a change in Mobility itself (just a plugin, albeit an important one), I don't want to release a minor version for it.

@pwim I believe you're using dirty tracking in production, right? Maybe you could run this change through your tests? It will technically impact earlier versions as well as Rails 6 since it's an entire rewrite of the dirty plugin.

spec/mobility/plugins/active_model/dirty_spec.rb
```diff
@@ -179,11 +183,7 @@ def save
         expect(instance.title_was).to eq("foo")
 
         Mobility.locale = :fr
-        if ENV['RAILS_VERSION'].present? && ENV['RAILS_VERSION'] < '5.0'
-          expect(instance.title_changed?).to eq(nil)
-        else
-          expect(instance.title_changed?).to eq(false)
-        end
+        expect(instance.title_changed?).to eq(false)
```
shioyama: The fact that `title_changed?` returned `nil` in AR 4.2 was actually a bug. Rather than mimic the bug, Mobility will now just return `false` as in later versions. This is technically a behaviour change but since 4.2 was never really fully supported, I don't intend to try to do anything about it.

PR #346, opened 2019-10-21
Fix Dirty plugin with Rails 6 (0-8-stable branch)

shioyama:
This is a patch to the 0-8-stable branch of #343. The branches are far apart now, so this really has to be the last major change to 0.8.x.


shioyama:
Wrong target branch...

PR #348, opened 2019-10-21
Fix issues with dirty plugin by dynamically defining dirty methods

shioyama:
Resolving issues raised in #347, for master branch. For a summary see [here](https://github.com/shioyama/mobility/pull/347#issuecomment-544742401), and 011575d9859b086330197d50427773e79f29d94d.

### DIFF
--- a/lib/mobility/plugins/active_model/dirty.rb
+++ b/lib/mobility/plugins/active_model/dirty.rb
@@ -34,15 +34,11 @@ The following methods are also patched to work with translated attributes:
         # Builds module which adds suffix/prefix methods for translated
         # attributes so they act like normal dirty-tracked attributes.
         class MethodsBuilder < Module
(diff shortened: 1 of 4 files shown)


shioyama:
@pwim This should roughly do it, minus a few more clean up changes. This is still on `master`, I'll update the other PR once this is passing and working.

As described earlier, naming methods on the tracker carefully pays off because we can dynamically define all the attribute methods using introspection and just delegate to the tracker now. Seems to work ok in all versions, and should quietly support future changes as well.

shioyama:
Ok, squashed commits and added more test coverage. Once specs pass, I'll merge and update (or maybe replace) #347, then we can finally release this as 0.8.9.

shioyama:
Merged! Now just need to rebase 0-8-stable...

PR #350, opened 2019-10-24
Fix Dirty plugin with Rails 6 (0-8-stable branch), take 2

shioyama:
Second attempt, following #347


shioyama:
Arg, keep forgetting to make the PR to 0-8-stable...

PR #351, opened 2019-10-24
Fix Dirty plugin with Rails 6 (0-8-stable branch), take 2

shioyama:
Second attempt, following #347

### DIFF
--- a/spec/mobility/plugins/active_record/dirty_spec.rb
+++ b/spec/mobility/plugins/active_record/dirty_spec.rb
@@ -221,13 +221,13 @@ describe "Mobility::Plugins::ActiveRecord::Dirty", orm: :active_record do
           expect(article.saved_change_to_title?).to eq(true)
           expect(article.saved_change_to_title).to eq([nil, "foo"])
           expect(article.title_before_last_save).to eq(nil)
-          #expect(article.title_in_database).to eq("foo")
+          expect(article.title_in_database).to eq("foo")
 
           # attribute handlers
           expect(article.saved_change_to_attribute?(:title_en)).to eq(true)
-          #expect(article.saved_change_to_attribute(:title_en)).to eq([nil, 'foo'])
-          #expect(article.attribute_before_last_save(:title_en)).to eq(nil)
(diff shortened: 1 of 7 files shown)


shioyama:
This includes #349, which I'll merge first once specs pass there.

spec/mobility/plugins/active_record/dirty_spec.rb
```diff
@@ -233,22 +239,31 @@ def clear_changes_information
...
+            expect(article.title_change_to_be_saved).to eq(nil)
+            expect(article.title_in_database).to eq("bar")
+
+            # attribute handlers
+            expect(article.saved_change_to_attribute?(:title_en)).to eq(true)
+            #expect(article.saved_change_to_attribute(:title_en)).to eq(['foo', 'bar'])
+            #expect(article.attribute_before_last_save(:title_en)).to eq('foo')
+            #expect(article.will_save_change_to_attribute?(:title_en)).to eq(false)
+            #expect(article.attribute_change_to_be_saved(:title_en)).to eq(nil)
+            #expect(article.attribute_in_database(:title_en)).to eq('bar')
```
shioyama: @pwim These specs all currently fail on 0-8-stable, so these other attribute methods (`saved_change_to_attribute`, etc.) don't work with translated attributes, just `saved_change_to_attribute?`. I'm not exactly sure why but the fact that they worked or not was pretty much dependent on internals of AR/AM.

These other specs will now *pass* with the new changes (once I commit them here), and they will be independent of those AM/AR internals.

shioyama:
Ok everything is up to date here now. If/when specs pass, I will merge to the 0-8-stable branch and we can finally release this.

In the end, this actually does quite a lot more than the dirty plugin did previously.

pwim:
Cool. I ran it against my application's specs and they all pass now.

shioyama:
Thanks for checking! I'll need to cherry-pick #353 in here as well before merging/releasing.

shioyama:
Ok cherry-picked last commit from master. Once this one passes, we should finally be good to release.

PR #353, opened 2019-10-25
Ensure that public/private scope of changes_applied, etc do not change

shioyama:
Previously, we were looking at the Active*Model* methods to se if they were private or public, and defining overrides accordingly. However, in some versions of Rails, the methods can be private in one module, but public in the other... so we need to ensure that the code actually looks at the correct +dirty_class+ to determine if the methods should be
public or private.

I never thought this dirty fix would be so complex...


shioyama:
Merged in ad9c66d, accidentally cherry-picked it so this wasn't automatically merged/closed. I'll just close it now.

PR #357, opened 2020-01-17
Ruby 2.7.0

f1sherman:
Fix specs and deprecations in Ruby 2.7.0. Also add Ruby 2.7 to Travis config.

I'm pretty new to this codebase, my apologies if I missed anything!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,12 @@ matrix:
       env: DB=mysql ORM=active_record RAILS_VERSION=6.0
     - rvm: 2.4
       env: DB=sqlite3 ORM=active_record RAILS_VERSION=6.0
+    - rvm: 2.7
+      env: DB=postgres ORM=active_record RAILS_VERSION=4.2
+    - rvm: 2.7
+      env: DB=mysql ORM=active_record RAILS_VERSION=4.2
+    - rvm: 2.7
+      env: DB=sqlite3 ORM=active_record RAILS_VERSION=4.2
 
 before_script:
   - bundle exec rake db:create db:up
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ rvm:
   - 2.4
   - 2.5
   - 2.6
+  - 2.7
 env:
   - DB=postgres ORM=active_record RAILS_VERSION=4.2
   - DB=mysql ORM=active_record RAILS_VERSION=4.2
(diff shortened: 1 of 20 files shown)


f1sherman:
I'm having a hard time interpreting the build failures. There are several of them and for all of the ones I've looked at AFAICT they don't seem likely to have been caused by these changes. Let me know if there is something I can do to help get those resolved!

shioyama:
Thanks for the PR! Build failures may be a compatibility issue with dependencies, I'll check later.

f1sherman:
Sounds great, thank you!

shioyama:
Just glanced now but seems like the errors are from some Thor deprecation, should be easy to fix. I'll try to find time to have a look today.

shioyama:
Created #363 to fix Thor failures in specs, I'll try to get that merged so we have a clean slate before evaluating this one.

Thanks again for the PR! I'll try to follow up on this asap.

f1sherman:
Sounds great, thank you!

f1sherman:
@shioyama looks like everything is passing now except for performance tests (not sure why these are failing) and an error in the postgres/sequel 4 test job (maybe we should configure that one to be skipped?). Let me know if there are any other changes you'd like me to make!

shioyama:
Thanks! Actually the failing performance spec means something, it's failing because using **options creates an extra hash object, something ideally I'd.like to avoid... need to research a bit.

n-rodriguez:
Hi there! Any news?

shioyama:
Ok I did a bit of research on this, and it seems that the wonderful @jeremyevans has made a change to Ruby which [reduces allocations for keyword argument hashes](https://github.com/ruby/ruby/pull/2945). This should fix the issue here with the failing test since it will allow using `**options` without allocating extra objects.

I know the ruby warnings are annoying, but I'd like to wait until that change is released to merge this, otherwise we'll forget about it and never get back the original performance we had. The change was merged to Ruby in March so I assume it will be included in a release soon.

shioyama:
Another update: it seems that these warnings may soon be removed. So let's wait and see what happens.

https://discuss.rubyonrails.org/t/new-2-7-3-0-keyword-argument-pain-point/74980

f1sherman:
Looks like the warnings will be removed in Ruby 2.7.2 but the behavior will still change in Ruby 3.0:
https://discuss.rubyonrails.org/t/new-2-7-3-0-keyword-argument-pain-point/74980/47

shioyama:
Updated this in #460, let's move the conversation there. I'm down to only one warning, but perf specs will fail. I think that's ok though.

PR #364, opened 2020-03-11
Fix fallthrough accessor method_missing not passing all options to super

doits:
When mobility handled something it wasn't responsible for, it didn't pass all options to `super` and actually swallowed `options`. Other method missing handlers couldn't work as intended then.

### DIFF
--- a/lib/mobility/plugins/fallthrough_accessors.rb
+++ b/lib/mobility/plugins/fallthrough_accessors.rb
@@ -53,7 +53,7 @@ model class is generated.
             locale = "#{locale}-#{suffix.upcase}" if suffix
             public_send("#{attribute}#{$4}", *arguments, **options, locale: locale.to_sym)
           else
-            super(method_name, *arguments, &block)
+            super(method_name, *arguments, **options, &block)
           end
         end
 
--- a/spec/mobility/plugins/fallthrough_accessors_spec.rb
+++ b/spec/mobility/plugins/fallthrough_accessors_spec.rb
@@ -16,6 +16,23 @@ describe Mobility::Plugins::FallthroughAccessors do
     it_behaves_like "locale accessor", :title, :en
     it_behaves_like "locale accessor", :title, :de
     it_behaves_like "locale accessor", :title, :'pt-BR'
+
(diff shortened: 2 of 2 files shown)


doits:
Oh and just one more idea (not yet fully tested and not thought to the end): It might be possible to speed up fallthrough accessor by defining an instance method *inside* `method_missing`, something like this:

```rb
        define_method :method_missing do |method_name, *arguments, **options, &block|
          if method_name =~ method_name_regex
            attribute = $1.to_sym
            locale, suffix = $2.split('_')
            locale = "#{locale}-#{suffix.upcase}" if suffix

            singleton_class.define_method(method_name) do |*arguments, **options|
              public_send("#{attribute}#{$4}", *arguments, **options, locale: locale.to_sym)
            end

            public_send(method_name, *arguments, **options)
          else
            super(method_name, *arguments, **options, &block)
          end
        end
```

This way `method_missing` will handle it only once and define the instance method. Next calls to this method will not go through `method_missing` anymore.

One has to think about how the method is defined exactly (especially the arguments), but this could be a way to speed up fallthrough accessors which will then have only a penalty on first call. Though this is something for another PR.

doits:
@shioyama maybe you can find some time to look at this? It is a fairly small change that unbreaks my App.

shioyama:
Sorry I will try to get to this and other issues in the next week or two.

shioyama:
Sorry for the delay! Thanks for the change, it looks great :+1: Do you think you could add a test so we avoid any future regression?

shioyama:
Never mind I added a spec in ￼abd67a2. If specs pass I'll merge & release :slightly_smiling_face:

doits:
Thanks!

shioyama:
It's passing :tada: I'll merge it and deploy later tonight, thanks for catching this!

shioyama:
This has been merged in b0b5d91eb58436eec55665b34f282faa14ab04da (master branch) and b0b5d91eb58436eec55665b34f282faa14ab04da (0-8-stable branch). Closing.

shioyama:
Released in [0.8.12](https://rubygems.org/gems/mobility/versions/0.8.12) :tada:

PR #369, opened 2020-05-19
Define default backend and options on Attributes class

shioyama:
The objective here is to remove `Mobility::Configuration` and instead define defaults in a subclass of `Mobility::Attributes`. This is more transparent and allows further nesting of configuration through inheritance.

Usage:

```ruby
  backend :key_value
  options cache: true, presence: false, query: true, fallbacks: false
end
```

The options and backend can still be overridden in each instance:

```ruby
class Post
  include TranslatedAttributes.new(:title, backend: :table, fallbacks: true)
end
```

### DIFF
--- a/lib/mobility.rb
+++ b/lib/mobility.rb
@@ -169,15 +169,12 @@ module Mobility
     # (see Mobility::Configuration#default_backend)
     # @!method default_backend
 
-    # (see Mobility::Configuration#default_options)
-    # @!method default_options
-    #
     # (see Mobility::Configuration#plugins)
     # @!method plugins
     #
     # (see Mobility::Configuration#default_accessor_locales)
     # @!method default_accessor_locales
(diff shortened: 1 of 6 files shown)


shioyama:
Thinking maybe this would be better:

```ruby
  plugin :cache, default: true
  plugin :presence, default: true
  plugin :query, default: true
  plugin :fallbacks, default: true

  default_backend :key_value
end
```

So you declare each plugin along with its default option value, along with (optionally) the default backend.

```ruby
class Post
  include TranslatedAttributes.new(:title)
end
```

This would use the default backend and default options, or you could override them:

```ruby
class Post
  include TranslatedAttributes.new(:title, fallbacks: false, presence: false)
end
```

Of course this way of declaring translations will be wrapped in `translates` as in 0.8.x, but behind the scenes I want the actual core API to be simpler with less magic hiding what's actually happening. Either way will be supported but as of 1.x the format above will be the one used mainly in docs, etc.

PR #373, opened 2020-05-22
Deprecate Ruby 2.4, support Ruby 2.7

shioyama:
2.4 was EOLed on 2020-03-31.


shioyama:
Oh that's unexpected, some issues on 2.7...

shioyama:
This looks like it's green but Travis is not actually catching Github webhooks, so tests have not actually been run yet...

shioyama:
Rebased and updated on master to use Github Actions instead of Travis.

PR #374, opened 2020-05-23
Fix issue with method_missing in fallthrough_accessors

### LINKED ISSUE #376: 0.8.12 breaks method_missing attribute handlers in ActiveModel/ActiveRecord
Issue caused by #364, where the `method_missing` override in the FallthroughAccessors plugin was changed to pass `**options` up to `super`. Problem is that keyword arguments are not really expected in a call to `method_missing`, and so `ActiveModel::AttributeMethods` chokes on it.

## Context

Notes from Gitter conversation with @Bramjetten:

image

## Possible Fix

See: #374

shioyama:
An issue was raised in gitter chat about options keywords being incorrectly passed to `method_missing` in ActiveModel/ActiveRecord with the change in #364.

The problem is subtle but I've added a test and a tentative fix for it here.

Fixes #376.

### DIFF
--- a/lib/mobility/plugins/fallthrough_accessors.rb
+++ b/lib/mobility/plugins/fallthrough_accessors.rb
@@ -49,14 +49,18 @@ model class is generated.
       def define_fallthrough_accessors(*names)
         method_name_regex = /\A(#{names.join('|')})_([a-z]{2}(_[a-z]{2})?)(=?|\??)\z/.freeze
 
-        define_method :method_missing do |method_name, *arguments, **options, &block|
+        define_method :method_missing do |method_name, *args, &block|
           if method_name =~ method_name_regex
-            attribute = $1.to_sym
+            attribute_method = "#{$1}#{$4}"
             locale, suffix = $2.split('_')
             locale = "#{locale}-#{suffix.upcase}" if suffix
-            public_send("#{attribute}#{$4}", *arguments, **options, locale: locale.to_sym)
+            if $4 == '=' # writer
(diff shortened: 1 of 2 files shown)


shioyama:
Ok Travis no longer running tests...

Bramjetten:
It fixed the previous bug, but it introduced two new ones for me. 😅

Having a default doesn't work anymore:
```ruby
translates :title, fallbacks: true
translates :menu_title, :seo_title, default: -> { title }
```

Schermafbeelding 2020-05-23 om 10 16 32
(This should be a list of menu_titles, which fallback to the title attribute when the actual menu_title  attribute isn't set)

Also, when trying to save any translated model (page or product in my case) I get this error:

Schermafbeelding 2020-05-23 om 10 16 13

shioyama:
Oh wait, sorry this is a PR from the `master` branch, which is quite different from 0.8.x. I'll cherry-pick to 0-8-stable so you can try the change there instead.

shioyama:
Also, note to self: the failing test is not actually failing. Need to fix.

shioyama:
Ok, so the failing test added here fails in Ruby versions < 2.7, but passes in 2.7 because there's a change in how keyword arguments are handled.

shioyama:
Restarting with two new PRs for master and 0-8-stable branches.

shioyama:
Re-opening, now running tests on GA. 0-8-stable fix is in #384.

shioyama:
Sorry never mind will use #377.

PR #380, opened 2020-05-25
Update Gemfile to work with Github Actions

shioyama:
Get this to work.


shioyama:
Ok after a lot of trial and error, finally almost there. I'll clean this up and we'll switch to github actions later this week.

PR #387, opened 2020-05-31
Extract reader/writer and backend setup into plugins

shioyama:
It occurred to me the other day that the setup stuff for readers, writers and the backend itself in `Mobility::Attributes` could themselves be extracted into Plugins, so this PR does this. And, amazing thing, IT WORKS!

I find this really remarkable. Everything is pretty much made up of plugins now. The backend plugin is a bit special, mind you, as it needs to be included first, and I need to think a bit how to avoid confusion about this one, but what's nice is that all the stuff that is backend specific is extracted into `Mobility::Plugins::Backend`, and what's left in `Mobility::Attributes` is exclusively to do with the attributes themselves, and not the backend.

The new format for configuring Mobility will look something like this (for a config with the table backend as default, etc):

```ruby
  plugin :backend, default: :table
  plugin :accessor
  plugin :fallbacks
  plugin :cache
  plugin :dirty
end
```

I think there could be a default subclass of `Mobility::Attributes` which includes `backend` and `accessor` to start, since 99% of applications will want those. But the nice thing is that by separating all these things, you could have (for example) a backend defined on the model, but define your own reader and writer methods to access that backend, and still use all the other plugins (like fallbacks, etc) with no issue. Or, you could use locale accessors but define your own accessor methods with whatever translation fetching logic you want, without ever even defining backends or anything else (`title_en` would just map to `title(locale: :en)` and your accessor method would have to handle that).

It also makes testing cleaner since stuff can be tested in total isolation.


lib/mobility/attributes.rb
```diff
@@ -116,52 +116,16 @@ def self.plugin(name)
...
-
-        klass.include InstanceMethods
-        klass.extend ClassMethods
-
-        backend_class.setup_model(klass, names)
-
-        backend_class
-      end
+      klass.extend ClassMethods
+      nil
```
shioyama, line 128: Note to self: `nil` is bad here, this is a placeholder for something better. :thinking:

shioyama:
If tests pass, I do believe this will be ready to merge :crossed_fingers:

PR #393, opened 2020-06-11
Register plugins explicitly

shioyama:
This is an idea borrowed from Shrine, Roda and other gems. Basically, we shouldn't be just assuming every plugin is in the Mobility::Plugins namespace, and that everything in this namespace is a plugin (some classes like Mobility::Plugins::ActiveModel, etc are actually not).

By explicitly declaring the mapping between keys and plugin modules, things are clearer, easier to test, and easier to extend (e.g. if you don't require a plugin from Mobility, you can use the same key for your own plugin module).

### DIFF
--- a/lib/mobility/plugins/attribute_methods.rb
+++ b/lib/mobility/plugins/attribute_methods.rb
@@ -34,5 +34,7 @@ attributes only.
         model_class.include module_builder.new(*attribute_names)
       end
     end
+
+    register_plugin(:attribute_methods, AttributeMethods)
   end
 end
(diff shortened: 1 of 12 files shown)


lib/mobility/plugins/attribute_methods.rb
```diff
@@ -34,5 +34,7 @@ def include_attribute_methods_module(model_class, *attribute_names)
         model_class.include module_builder.new(*attribute_names)
       end
     end
+
+    register_plugin(:attribute_methods, AttributeMethods)
```
shioyama, line 38: Thinking this module should really go in v1.0, it's very AR/Rails specific. Maybe a separate gem or something, or just a note in the wiki with some copy-paste code since it's so small anyway.

PR #404, opened 2020-07-03
Extract Attributes into its own plugin

shioyama:
That's it, now basically everything is a plugin.

Also, previously when calling methods like `mobility_attributes`, we'd `grep` all modules which are instances of `Mobility::Attributes`, grab their `names` and merge/flatten them. Same idea for backend classes. With this PR, we instead keep a running copy of what they are in an array, and add to it. Same for backend classes except we merge (since backend classes is a hash of keys->class).

Also: removed `BackendCache`, we don't need it.

### DIFF
--- a/lib/mobility.rb
+++ b/lib/mobility.rb
@@ -29,7 +29,6 @@ module Mobility
   class Error < StandardError
   end
 
-  require "mobility/attributes"
   require "mobility/backend"
   require "mobility/backends"
   require "mobility/configuration"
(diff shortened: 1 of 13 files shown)


shioyama:
In passing, I'll mention the only place I could find that actually explains the return values of `block.parameters` in any detail is [this page in Japanese](https://docs.ruby-lang.org/ja/latest/method/Proc/i/parameters.html).

PR #407, opened 2020-07-05
Improve ORM delegation

shioyama:
Lots of work to do here. Let's see what we can do...


shioyama:
Going to try this again differently.

Issue #410, opened 2020-07-08
undefined method `default_backend=' for Mobility::Configuration


sedubois:
## Context

When updating Mobility from 0.8.13 to master (2cffe9e6fe8da1d5c443eb2ecad87d64ee182d8c), I get this error on startup:

```
=> Booting Puma
=> Rails 6.0.3.2 application starting in development 
=> Run 'rails server --help` for more startup options
Exiting
Uncaught exception: undefined method 'default_backend=' for #<Mobility::Configuration:0x00007fe7f092fe20>
Did you mean?  default_backend
	/Users/sdubois/Repos/webapp/config/initializers/mobility.rb:4:in 'block in <main>'
	/Users/sdubois/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/bundler/gems/mobility-2cffe9e6fe8d/lib/mobility.rb:185:in 'configure'
	/Users/sdubois/Repos/webapp/config/initializers/mobility.rb:1:in '<main>'
```

The failure happens on the configuration line `config.default_backend = :table` in the initializer (which I did not change).

## Expected Behavior

Application initialization works as before.

## Actual Behavior

Application initialization fails.

## Possible Fix

Appears to be a regression from https://github.com/shioyama/mobility/commit/07a400ec9e27a7c02503028670a42a080676286d#diff-26e809a7ee85d202203e4df2e1098f2cL55 (`attr_accessor :default_backend` was removed).

It might also help to add a non-regression test?

shioyama:
Yeah configuration will change quite a bit in v1.0, that's why I added a note to the readme. I'm trying to maintain rough maps so that simple stuff will still work, but it's not done yet.

I'm going to close this as a known issue (documentation). It's not broken, it's an (undocumented) API change. Thanks for trying master though! I will try to add some info to the wiki soon about how configuration and some other things will work in v1.0 soon.

shioyama:
And you're right: the initializer needs to be updated. And you're right that there should be a test for the initializer so that it stays in sync. Good points! Maybe you could create an issue about having a test that the initializer works?

PR #412, opened 2020-07-08
Add text key-value backend compatibility for ActionText

### LINKED ISSUE #385: Rails ActionText backend to translate rich text
I would like the `:key_value` backend to be configurable in order to support translating ActionText rich text.

## Context

It [appears](https://github.com/globalize/globalize/issues/753) Mobility is becoming the new "Rails I18n de-facto standard library for ActiveRecord model/data translation", to re-use Globalize's description 😉 I also guess that many apps, including my own bilingual content-centric app, use rich text rather than plain text, where ActionText is supposed to be the ad-hoc Rails solution. Additionally, Mobility describes itself as adapting to different storage needs. I would thus like Mobility and ActionText to work elegantly together.

## Expected Behavior

Being able to store and query translated ActionText rich text using a single table join using the existing Mobility API.

## Actual Behavior

I am not aware of a way to make ActionText and Mobility work together without requiring two levels of joining. [This example](https://github.com/rails/actiontext/issues/32#issuecomment-450653800) shows that translating ActionText can be done, however it requires two nested joins and fragile patches of code.

## Possible Fix

Similar to the `:key_value` backend, ActionText also works by establishing a polymorphic relation on the model. This raises an opportunity to translate ActionText rich text "for free", i.e. at no extra performance cost compared to untranslated rich text, by directly using the `action_text_rich_texts` table as a `:key_value` translation backend instead of the `mobility_text_translations` table.

Consider the Mobility and ActionText schemas, they already have a 1:1 mapping, except for the `locale` column: `key <=> name`, `value <=> body`, `translatable_id <=> record_id`, `translatable_type <=> record_type`:

<details>

```rb
create_table "mobility_text_translations", force: :cascade do |t|
  t.string   "locale"
  t.string   "key"
  t.text     "value"
  t.integer  "translatable_id"
  t.string   "translatable_type"
  ...
end
```

```rb
create_table "action_text_rich_texts", force: :cascade do |t|
  t.string "name", null: false
  t.text "body"
  t.bigint "record_id", null: false
  t.string "record_type", null: false
  ...
end
```

</details>

Whereas for the example above, ActionText would normally store:
```rb
{ record_type: 'Post', record_id: 1, name: 'content', body: '<h1>hello world!</h1>' }
```

... it would instead store:
```rb
{ record_type: 'Post', record_id: 1, name: 'content', locale: 'en', body: '<h1>hello world!</h1>' }
{ record_type: 'Post', record_id: 1, name: 'content', locale: 'fr', body: '<h1>bonjour le monde !</h1>' }
```

So if Mobility offered a way to configure the table name and columns for the `:key_value` backend, in theory everything should "just work" 🤞😄

Mobility could be instructed to achieve it with something like this:
```rb
Mobility.configure do |config|
  config.default_backend = :key_value
  config.backends.key_value.text.table = :action_text_rich_texts
  config.backends.key_value.text.translatable = :record
  config.backends.key_value.text.key = :name
  config.backends.key_value.text.value = :text
end
```

Or with a shorthand:
```rb
Mobility.configure do |config|
  config.default_backend = :key_value
  config.key_value_text_backend = :action_text
end
```

The migration should be adapted to first check if the table exists, and if it exists, to insert the required columns/indices.

Then at the model level the following (existing) API should suffice:

```rb
# app/models/post.rb
  has_rich_text :content
  translates :content
end
```

__________

@shioyama I understand that you do not have much bandwidth, but would you consider a PR? In that case I would be willing to give it a try with your guidance.

sedubois:
@shioyama here is a start. I tried to base the work on the master branch but [as noted](https://github.com/shioyama/mobility/issues/410), I encountered issues. Could you provide some info to work with master instead of the 0.8 branch to avoid conflicts?

I took the code from [this POC](https://github.com/sedubois/mobility/commit/7ec6bdf71bc64c32a472eabd4db77ffccef0aa99) and made it configurable by passing options as follows:
```rb
Mobility.configure do |config|
  config.default_backend = :key_value
  # ...
  config.default_options[:type] = :text
  config.default_options[:class_name] = 'ActionText::RichText'
  config.default_options[:key_column] = :name
  config.default_options[:value_column] = :body
  config.default_options[:translatable] = :record
end
```

Is this the right approach? As it requires a lot of time to find my way around, I first would like to make sure I'm going in the right direction.

If so, I will continue adapting the code so that string translations continue to work as expected + replace other hard-coded changes with proper configuration.

Fixes #385


lib/mobility/backends/active_record/key_value.rb
```diff
@@ -155,6 +155,9 @@ def visit_default(_)
       setup do |attributes, options|
         association_name   = options[:association_name]
         translations_class = options[:class_name]
+        key_column         = options[:key_column]
```
sedubois, line 158: For some reason the auto-generated `key_column`, etc methods are not yet available here.

CUnknown: I am not experiencing this when using you branch to experiment with the solution.  When I put a debug statement on line 161, I will get this debugger during project boot, as expected and the `association_name`, `translations_class`, `key_column`, `value_column` and `translatable` variables are all filled with the values that I provided in the `mobility.rb` initializer. This looks to be fine!

shioyama:
> Could you provide some info to work with master instead of the 0.8 branch to avoid conflicts?

Sorry, `master` is a bit a work in progress, but `default_options` will no longer be the way to set default options. It was always a problematic pattern in Mobility and I've wanted to improve it, and v1.0 will be the moment to do so. `Mobility.default_backend=` will allso change.

I haven't explained this anywhere yet but let me just give it a shot here. (I just merged a couple PRs (#420, #419) to fill out some gaps.)

So basically, in 1.0, the idea is to take away all the weird tricks in Mobility and centralize everything around plugins. Plugins can have default values, which combined become what `default_options` are before 1.0.

So previously, you would do this:

```ruby
Mobility.configure do |config|
  config.default_backend = :key_value
  config.default_options[:locale_accessors] = [:en, :ja]
  config.default_options[:type] = :text
  # ...
end
```

There are a bunch of problems with this, but one obvious one is that options for _plugins_, like `locale_accessors` here, look exactly like options for _backends_, like `type` here.

The default way to configure Mobility is by subclassing `Mobility::Attributes`, and inside the subclass calling class methods to define plugins and their defaults, as well as other defaults not associated with a plugin (i.e. backend defaults):

```ruby
  plugin :backend,          default: :key_value
  plugin :locale_accessors, default: [:en, :ja]
  default :type, :text
  # ...
end
```

Notice that even the backend is a plugin now, with a default (which corresponds to the current `Mobility.default_backend`).

Now instead of `translates`, you would do this:

```ruby
  include TranslatedAttributes.new(:title)
end
```

(Actually this is also possible in 0.8.x, but nobody does it this way.)

I realize that this "including an instance of a thing" will be unintuitive, so this will mostly map back to the same way things are done in 0.8 with `translates`. But to configure, you won't use `default_options` or `default_backend` anymore, and instead use `plugin` and `default`, like this:

```ruby
Mobility.configure do |config|
  config.plugin :backend,          default: :key_value
  config.plugin :locale_accessors, default: [:en, :ja]
  default :type, :text
  # ...
end
```

In specs, there is a helper `plugin_setup` ([here](https://github.com/shioyama/mobility/blob/5accbbb04aedc7feb127cfa7686a566c898c161a/spec/support/helpers.rb#L144-L171)) which you can use to setup an attributes class with a given set of plugins. That's what I've been using in most specs in 1.0.

I'm not sure if that will help with the problems here, but since I haven't actually documented this anywhere yet I thought I'd just explain quickly. I plan to update all documentation once I've finished refactoring existing code to the new patterns.

shioyama:
Also, to add something which is probably confusing people. Previously, certain plugins were automagically included without even asking for them, see [here](https://github.com/shioyama/mobility/blob/b24eb429b9b4ecf8cd1c4e7dae47d2e40de4d23e/lib/mobility/configuration.rb#L108-L113) and [here](https://github.com/shioyama/mobility/blob/b24eb429b9b4ecf8cd1c4e7dae47d2e40de4d23e/lib/mobility/configuration.rb#L114-L124). So without doing anything, you could use the Query plugin, for example.

In 1.0, everything is explicit and nothing is enabled by default, so to use the query plugin you need to explicitly ask for it:

```ruby
Mobility.configure do |config|
  # ...
  config.plugin :query
  # ...
end
```

The same goes for `dirty`, previously you could just pass `dirty: true` to translates and it would work, but now you need to actually_enable_ the plugin first in the global configuration.

My intention here is to make everything much more intentional, so you only get what you actually ask for. I think in the long run this will make things much clearer and easier to debug/extend/etc

YassineDM:
Hello @sedubois, I was wondering if you managed to find a working solution because I am also very interested in the feature...

sedubois:
Hi @YassineDM, I updated my app to Mobility 1.0.0.beta1, so when I get the chance ill try to rebase the demo in this PR on that branch, and take it from there. Feel free to try as well, let’s see who is able to progress first and we can report our findings here.

YassineDM:
Thank you @sedubois for coming back to me. I am still on Mobility 0.8.13 but did not yet dig deeper into the topic. So I will let you know by then...

shioyama:
Very interested in this discussion, although I don't have time to contribute directly. However, I _can_ promise to help with any understanding of the changes in 1.0, since they are not entirely documented yet. Not sure how much they will affect the stuff here, but just let me know if something is unclear. It's not super obvious I know :sweat_smile:

lib/mobility/backends/active_record/key_value.rb
sedubois: In my latest tests I still needed these lines, so leaving them for now.

lib/mobility/backends/active_record/key_value.rb
```diff
@@ -204,12 +211,15 @@ def self.included(model_class)
...
-                where(translatable: self).destroy_all
+                where(translatable => self).destroy_all
             end
           end
         end
       end
 
+      # FIXME: replace:
+      #   - `:translatable` with `options[:translatable]`
+      #   - `:key` with `options[:key_column]`
```
sedubois, line 222: With my limited Ruby-fu I don't know how to declare this class in a way that respects the :translatable and :key_column options I added.

sedubois:
I've pushed the latest changes:
- rebased on master (Mobility 1.0.1)
- created ActionText backend class extending ActiveRecord::KeyValue

It currently allows this API (but could be nice to simplify e.g. to `has_translated_rich_text :content`):

```rb
  translates :content, backend: :action_text
  has_rich_text :content
end
```

+ applied these migrations: https://github.com/shioyama/mobility/issues/385#issuecomment-637082724.

I'm now trying to figure out why the locale isn't being set correctly in the backend and also not used to filter out the translations to query. I guess either the ActionText::RichText class should be patched or another one should be used.

lib/mobility/backends/active_record/key_value.rb
shioyama: In order to achieve this, we'd need to make this class and its subclasses dynamically generated in the `setup` block. This is possible and shouldn't be a problem since we're anyway subclassing these classes now.

shioyama:
@sedubois I took a bit of time to look into this. Although it's a really interesting idea, I don't think it's quite so simple, unfortunately. The problem is that Mobility defines attribute getter/setter methods, and so does ActionText. The two don't work with each other out of the box because they don't plug into each other (they define the same methods, e.g. `content`, `content?`, and `content=`).

Mobility can be configured to _not_ define those methods (simply remove `reader` and/or `writer` from the configuration), but the problem is that ActionText still needs to be patched to look for the text with the matching locale. I'm not sure this can be done so easily.

Just quickly looking, I found this is where `has_action_text` is defined:

https://github.com/rails/rails/blob/5262ec59002e461b2fdb8cb81f0ac1623b1cd2fd/actiontext/lib/action_text/attribute.rb#L29-L49

```ruby
      def has_rich_text(name)
        class_eval <<-CODE, __FILE__, __LINE__ + 1
          def #{name}
            rich_text_#{name} || build_rich_text_#{name}
          end

          def #{name}?
            rich_text_#{name}.present?
          end

          def #{name}=(body)
            self.#{name}.body = body
          end
        CODE

        has_one :"rich_text_#{name}", -> { where(name: name) },
          class_name: "ActionText::RichText", as: :record, inverse_of: :record, autosave: true, dependent: :destroy

        scope :"with_rich_text_#{name}", -> { includes("rich_text_#{name}") }
        scope :"with_rich_text_#{name}_and_embeds", -> { includes("rich_text_#{name}": { embeds_attachments: :blob }) }
      end
```

Now, _maybe_ Mobility could be made to hook into this stuff somehow, but it wouldn't be trivial. I'm happy to help but I have no knowledge of ActionText...

sedubois:
@shioyama I will see if I can define our own `has_translated_rich_text` which would combine the needs of ActionText and Mobility.

shioyama:
Actually, I'm not totally sure I'm following everything here, so take my comments with a grain of salt.

shioyama:
I think Mobility needs to define the getter and setter, so translates should come after has_action_text.

spec/mobility/backends/active_record/action_text_spec.rb
```diff
@@ -0,0 +1,510 @@
+require "spec_helper"
+
+return unless defined?(ActiveRecord)
+
+describe "Mobility::Backends::ActiveRecord::ActionText", orm: :active_record, type: :backend do
```
sedubois, line 5: This is a copy of `key_value_spec.rb`, where some specs which are not relevant to this scenario are commented out (e.g. when combining string and text types, as there is no need to configure such option in this case).

lib/mobility/backends/active_record/key_value.rb
```diff
@@ -210,6 +210,7 @@ module DestroyKeyValueTranslations
         def self.included(model_class)
           model_class.after_destroy do
             [:string, :text].each do |type|
+              # FIXME: undefined local variable or method `translatable'
               Mobility::Backends::ActiveRecord::KeyValue.const_get("#{type.capitalize}Translation").
                 where(translatable => self).destroy_all
```
sedubois: How to make options[:translatable] available here?

sedubois:
@shioyama how do I load `ActionText` in both the backend code and in the specs? Currently I get `undefined method `has_rich_text'`.

- when I add `include ::ActionText` to the backend I get `uninitialized constant ActionText`
- adding `require "action_text"` does not help

sedubois (commented):
369cf20561fe3bb4b8f765d20c1c546ed8b806ce is to fix record destruction in ActionText, but then for some reason a single after_destroy hook gets called even if there are several (e.g. if model has both string and text translations)?

sedubois (commented):
@shioyama with this change I'm now able to have translations in my app, just by adding `has_translated_rich_text :content` in the model. However I have no idea how to call this method in the tests. Also are you OK with this API?

sedubois:
Here is current status, @shioyama would be great to have your feedback:

### API

The functionalities of ActionText + Mobility are provided through a single new method `has_translated_rich_text`.

### Implementation

A new `ActionText` backend is defined as inheriting from the `ActiveRecord::KeyValue` backend (without using the `type` option). Initially I wondered if instead a new `:rich_text` type should be added directly to the KeyValue backend (i.e. it would handle `string`, `text`, and `rich_text`). However I felt like "type" is for customizing the value column type, whereas here the needs are broader: specific table name (`action_text_rich_texts`), specific key column name (`name`), specific value column name (`body`) and type (`text`) + extra code to integrate with ActionText (cfr. API mentioned above). So I thought it might make more sense to keep these considerations separated in a subclass.

In the end it seems the integration "just works" by doing the following (based on a limited check in my app, as I couldn't get specs to run properly yet):
- call ActionText's `has_rich_text`
- replace its `has_one :"rich_text_#{name}"` definiton to scope the locale
- apply Mobility's `translates`

### Blocking points (cfr. earlier comments)

- don't know how to get ActionText to load in the library/specs (but works in my app)
- don't know how to call `has_translated_rich_text` in specs
- after_destroy not properly called

lib/mobility.rb
```diff
@@ -95,6 +95,14 @@ def extended(model_class)
...
         include Mobility.translations_class.new(*args, **options)
       end
+
+      # FIXME move this elsewhere?
+      def model_class.has_translated_rich_text(name)
+        has_rich_text(name)
+        has_one :"rich_text_#{name}", -> { where(name: name, locale: I18n.locale) },
+                class_name: "ActionText::RichText", as: :record, inverse_of: :record, autosave: true, dependent: :destroy
+        translates name, backend: :action_text
+      end
```
shioyama, line 105: Yes this definitely does not belong here. I think the end goal should be to create a separate gem for this, since it's fairly specific.

However, we can make the changes in Mobility necessary to _allow_ that separate gem to work, e.g. making key/value columns configurable, etc.

lib/mobility.rb
```diff
@@ -95,6 +95,14 @@ def extended(model_class)
       def model_class.translates(*args, **options)
         include Mobility.translations_class.new(*args, **options)
       end
+
+      # FIXME move this elsewhere?
+      def model_class.has_translated_rich_text(name)
+        has_rich_text(name)
+        has_one :"rich_text_#{name}", -> { where(name: name, locale: I18n.locale) },
+                class_name: "ActionText::RichText", as: :record, inverse_of: :record, autosave: true, dependent: :destroy
```
shioyama, line 103: You should be able to add this to a `setup` block called from within `Mobility::Backends::ActionText`, like this:

```ruby
module Mobility
  module Backends
    class ActionText < ActiveRecord::KeyValue
      setup do |attributes, _options|
        attributes.each do |name|
          has_rich_text(name) }
          has_one :"rich_text_#{name}", -> { where(name: name, locale: I18n.locale) },
            class_name: "ActionText::RichText", as: :record, inverse_of: :record, autosave: true, dependent: :destroy
        end
      end
    end
  end
end
```

By defining it this way, you now only need to call `translates` and `has_rich_text` will be called for you (so no need for `has_translated_rich_text`.)

lib/mobility/backends/action_text.rb
```diff
@@ -0,0 +1,61 @@
...
+        # @!group Backend Configuration
+        # @option (see Mobility::Backends::KeyValue::ClassMethods#configure)
+        def configure(options)
+          raise ArgumentError, "The type option is unsupported with this backend." if options[:type]
+          options[:association_name] ||= "rich_text_translations"
+          options[:class_name]       ||= Translation
+          options[:key_column]       ||= :name
+          options[:value_column]     ||= :body
+          options[:translatable]     ||= :record
+          options[:table_alias_affix] = "#{model_class}_%s_#{options[:association_name]}"
```
shioyama, line 39: This was removed in https://github.com/shioyama/mobility/pull/486, you shouldn't need it anymore.

lib/mobility.rb
```diff
@@ -95,6 +95,14 @@ def extended(model_class)
       def model_class.translates(*args, **options)
         include Mobility.translations_class.new(*args, **options)
       end
+
+      # FIXME move this elsewhere?
+      def model_class.has_translated_rich_text(name)
+        has_rich_text(name)
+        has_one :"rich_text_#{name}", -> { where(name: name, locale: I18n.locale) },
```
shioyama, line 102: Shouldn't this be `Mobility.locale` ?

shioyama:
This is pretty interesting, but I think we need to think carefully about what we want to happen here, versus what is actually happening.

I ran this branch and noticed that although `translates` includes the `Mobility::Translations` module with `content` defined, ActionText defines `content` on the _class_ and this takes precedence. Setter/getter and presence (`content?`) are all defined in `has_rich_text` and those all take precedence over Mobility methods defined in the module.

The question is: do we even need to call `has_rich_text` at all? It defines getter/setters that anyway we want Mobility to override. It also defines these associations:

```ruby
        has_one :"rich_text_#{name}", -> { where(name: name) },
          class_name: "ActionText::RichText", as: :record, inverse_of: :record, autosave: true, dependent: :destroy

        scope :"with_rich_text_#{name}", -> { includes("rich_text_#{name}") }
        scope :"with_rich_text_#{name}_and_embeds", -> { includes("rich_text_#{name}": { embeds_attachments: :blob }) }
```

But the first one is already being overridden in `has_translated_rich_text` (which as I mentioned I think should be removed and instead implemented in the backend `setup` block).

So what if we just don't call `has_rich_text` at all and reproduce what that method does in the backend `setup` block?

Here's the entire method body again:

```ruby
      def has_rich_text(name)
        class_eval <<-CODE, __FILE__, __LINE__ + 1
          def #{name}
            rich_text_#{name} || build_rich_text_#{name}
          end

          def #{name}?
            rich_text_#{name}.present?
          end

          def #{name}=(body)
            self.#{name}.body = body
          end
        CODE

        has_one :"rich_text_#{name}", -> { where(name: name) },
          class_name: "ActionText::RichText", as: :record, inverse_of: :record, autosave: true, dependent: :destroy

        scope :"with_rich_text_#{name}", -> { includes("rich_text_#{name}") }
        scope :"with_rich_text_#{name}_and_embeds", -> { includes("rich_text_#{name}": { embeds_attachments: :blob }) }
      end
```

That seems straightforward to reproduce, the only thing the current code is missing are the last two scopes.

shioyama:
But actually, to continue my question about what we "want" here, what should the value of `content` actually be? In the case of ActionText, it's a record. In the case of Mobility it's a string value. I don't know ActionText but I assume that if it returns a string it will break lots of stuff.

The solution here might be to remove the `reader` and `writer` plugins (so tell Mobility not to define getter/setter methods) and instead define the getter/setter in the `setup` block directly on the class, using the backend (which you can get with `mobility_backends[attribute]`.

But I'm wondering really if there's much value in using Mobility at that point. It feels like a bit of a stretch honestly.

shioyama:
Actually looking more closely, it seems the setter would match what Mobility does. So it's really the getter that's the problem:

```ruby
def #{name}
  rich_text_#{name} || build_rich_text_#{name}
end
```

This defines a getter which returns an `ActionText::RichText` record.

You could do the same thing by redefining the `read` method on the `ActionText` backend like this:

```ruby
def read(locale, **options)
  translation_for(locale, **options)
end
```

So basically, you just return the record rather than fetching its value. I think this should work.

lib/mobility/backends/action_text.rb
```diff
@@ -0,0 +1,61 @@
...
+          super
+        end
+        # @!endgroup
+      end
+
+      # FIXME: replace:
+      #   - `:translatable` with `options[:translatable]`
+      #   - `:key` with `options[:key_column]`
+      class Translation < ::ActiveRecord::Base
```
shioyama, line 48: Make this inherit from `::ActionText::RichText`.

lib/mobility/backends/action_text.rb
```diff
@@ -0,0 +1,61 @@
...
+        self.table_name = "action_text_rich_texts"
```
shioyama, line 49: Once this inherits from `::ActionText::RichText`, this shouldn't be needed.

lib/mobility/backends/action_text.rb
```diff
@@ -0,0 +1,61 @@
...
+
+        belongs_to :record, polymorphic: true, touch: true
```
shioyama, line 51: Ditto to this.

shioyama:
Ok I pushed a couple commits which I think get this closer to where it might work. I probably won't have time to do much more, but the main thing I need is to know _what needs to change in Mobility to make this possible (or easier)_?

What I see right now:
- make `key` column configurable
- make `value` column configurable
- make `translatable` configurable

The first two are simple. I'm not really sure about the third one, but it _might_ be possible to deduce it via introspection on the translation class :thinking:

lib/mobility/backends/active_record/key_value.rb
shioyama: This shouldn't be necessary, since we're setting the `translation_class` to `::ActionText::RichText` (see my commits) and we thus never use this translation class.

lib/mobility/backends/action_text.rb
sedubois: Cfr. [earlier point](https://github.com/shioyama/mobility/pull/412#issuecomment-751447820): I get `uninitialized constant ActionText`

lib/mobility/backends/action_text.rb
```diff
@@ -42,14 +47,19 @@ def configure(options)
...
+
       # FIXME: replace:
       #   - `:translatable` with `options[:translatable]`
       #   - `:key` with `options[:key_column]`
-      class Translation < ::ActiveRecord::Base
-        self.table_name = "action_text_rich_texts"
-
-        belongs_to :record, polymorphic: true, touch: true
-
+      class Translation < ::ActionText::RichText
```
sedubois, line 62: As mentioned I wanted to do this since the beginning but in my tests it throws `uninitialized constant ActionText`. Also I have been writing those tests, even if using a non-final implementation, in the hopes of reaching a "green" state on the basis of which we can then clean up the mess.

shioyama: Ah sorry I didn't read carefully :sweat_smile: I also got that error which is why I added the `::`.

shioyama: Ah sorry I think the failures are because the tests only require activerecord, not all of rails.

I think it's fine to ignore the spec failures since this backend should really be a separate gem which can then just require actiontext and test with that loaded.

sedubois:
@shioyama how about moving mobility and related gems (mobility-ransack, mobility-actiontext) to their own GitHub org, that way they could be maintained together?

shioyama:
>  how about moving mobility and related repos (mobility-ransack, mobility-actiontext) to their own GitHub org, that way they could be maintained together?

That's definitely something I've considered, but I would only really do that when Mobility has more users and a broader base of contributors, otherwise I feel it would be a one-person org with only a few repos.

For this gem, I'd be more than happy if you wanted to create/author/maintain it, and then if an org is created in the future repos could be moved there.

sedubois:
@shioyama OK let me know when Mobility 1.1.0 is published so I can try to make a gem (never done that before).

sedubois:
@shioyama I'm thinking of using [`mobility-actiontext`](https://github.com/sedubois/mobility-actiontext) instead of `mobility-action_text` for the gem name, to match the Action Text gem name which is `actiontext`. What do you think?

Next I need to figure out how to actually write tests by loading Rails in the test suite (it seems this is not done in any Mobility test, correct?). Currently if I just `require "action_text` I get:

```
$ rake test
/Users/sdubois/Repos/mobility-actiontext/lib/mobility/backends/action_text.rb:58:in `<class:ActionText>': uninitialized constant ActionText::RichText (NameError)
```

shioyama:
>  I'm thinking of using `mobility-actiontext` instead of `mobility-action_text` for the gem name, to match the Action Text gem name which is actiontext. What do you think?

Sure :+1: 

> Next I need to figure out how to actually load Rails in the test suite (it seems this is not done in any Mobility test, correct?). 

It is if the `FEATURE` environment variable is set to `rails`, see:

https://github.com/shioyama/mobility/blob/b187ad45b5c75c1524c56fd69bdbc222b619e76f/Gemfile#L38-L41

and:

https://github.com/shioyama/mobility/blob/b187ad45b5c75c1524c56fd69bdbc222b619e76f/spec/spec_helper.rb#L11

sedubois:
> and:
> 
> https://github.com/shioyama/mobility/blob/b187ad45b5c75c1524c56fd69bdbc222b619e76f/spec/spec_helper.rb#L11

@shioyama this is what I tried:

```rb
# mobility-actiontext.gemspec
Gem::Specification.new do |spec|
  # ...
  spec.add_dependency "mobility", "~> 1" # to be replaced with 1.1 when available
  spec.add_dependency "rails", "~> 6.0"
end

# test/test_helper.rb
require "rails"
require "action_text"
require "mobility"
require "mobility/action_text"
```

However I still have:
```
$ rake test
/Users/sdubois/Repos/mobility-actiontext/lib/mobility/backends/action_text.rb:58:in `<class:ActionText>': uninitialized constant ActionText::RichText (NameError)
        from /Users/sdubois/Repos/mobility-actiontext/lib/mobility/backends/action_text.rb:29:in `<module:Backends>'
```

`ActionText::RichText` is defined in `models/action_text/rich_text.rb`, I just don't know how to load it.

shioyama:
`ActionText::RichText` is not autoloaded when you `require "action_text"`, so if I do this in in any Rails application I get the same:

```ruby
require 'action_text'
=> true
iActionText::RichText
Traceback (most recent call last):
        1: from (irb):2
NameError (uninitialized constant ActionText::RichText)
```

Nothing to do with Mobility or that backend code.

It seems that `ActionText::RichText` lives in `actiontext/app/models/action_text/rich_text.rb` which is not autoloaded, it's actually part of an engine.

It seems that if you have a Rails app and you comment out the line `require "action_text/engine"`, then `ActionText::RichText` is loaded (or can be loaded).

But if you _don't_, and then from console try `require "action_text/engine"`, this will not work and the constant will still not be loaded. So this is something to do with how engines are setup, and to be honest my knowledge of this part of Rails is pretty shallow...

You might need to setup a dummy app to do the testing... this seems to be how Rails does at least some of the tests there.

sedubois:
Thanks @shioyama, I came to the same conclusion. I've added a demo app and now tests are passing:

https://github.com/sedubois/mobility-actiontext

One last mystery is why GitHub CI tests are not working, although it runs locally 😔

sedubois:
Thanks @shioyama, I came to the same conclusion. I've added a demo app and now tests are passing:

https://github.com/sedubois/mobility-actiontext

After much struggle I also managed to make tests green in CI, however:
- `bundle exec rake test` doesn't work, I had to replace with `rake test` (don't know if that has any consequences)
- `rake test` generates apparently non-blocking warnings: 
<details>

```
/opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/rubocop-rails-2.9.1/lib/rubocop/cop/mixin/enforce_superclass.rb:7: warning: method redefined; discarding old included
/opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/rubocop-1.8.1/lib/rubocop/cop/mixin/enforce_superclass.rb:15: warning: previous definition of included was here
/opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/rubocop-rails-2.9.1/lib/rubocop/cop/mixin/enforce_superclass.rb:19: warning: method redefined; discarding old on_class
/opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/rubocop-1.8.1/lib/rubocop/cop/mixin/enforce_superclass.rb:27: warning: previous definition of on_class was here
/opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/rubocop-rails-2.9.1/lib/rubocop/cop/mixin/enforce_superclass.rb:25: warning: method redefined; discarding old on_send
/opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/rubocop-1.8.1/lib/rubocop/cop/mixin/enforce_superclass.rb:33: warning: previous definition of on_send was here
/opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/bundler-2.2.5/lib/bundler/definition.rb:32:in `build': /home/runner/work/mobility-actiontext/mobility-actiontext/test_app/Gemfile not found (Bundler::GemfileNotFound)
	from /opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/bundler-2.2.5/lib/bundler.rb:194:in `definition'
	from /opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/bundler-2.2.5/lib/bundler.rb:142:in `setup'
	from /opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/bundler-2.2.5/lib/bundler/setup.rb:20:in `block in <top (required)>'
	from /opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/bundler-2.2.5/lib/bundler/ui/shell.rb:136:in `with_level'
	from /opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/bundler-2.2.5/lib/bundler/ui/shell.rb:88:in `silence'
	from /opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/gems/3.0.0/gems/bundler-2.2.5/lib/bundler/setup.rb:20:in `<top (required)>'
	from <internal:/opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
	from <internal:/opt/hostedtoolcache/Ruby/3.0.0/x64/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
```

</details>

Any idea what might cause those?

`bundle install` was putting a native nokogiri darwin build in the Gemfile.lock (`1.11.1-x86_64-darwin`) but this is incompatible with the Ubuntu test machine, I workarounded that by copy-pasting the way nokogiri appears in my own app (`1.11.1`). I'm sure that's not the correct way though.

I must say that doing all this work was surprisingly painful and I would probably not have done it if I knew how much struggle it would be 😄

sedubois:
Thanks @shioyama, I came to the same conclusion. I've added a demo app and CI tests are now green:
https://github.com/sedubois/mobility-actiontext/actions/runs/485895295

NB: `bundle install` was putting a native nokogiri darwin build in the Gemfile.lock (`1.11.1-x86_64-darwin`) but this is incompatible with the Ubuntu test machine. I workarounded that by copy-pasting from my own app where it appears without architecture (`1.11.1`), maybe for historical reasons. I'm sure that's not the correct way to fix this though, as every time the Gemfile gets regenerated it adds back the darwin architecture.

Making the gem was a surprising amount of work, might not have done it if I knew at the outset 😄

sedubois:
Closing in favor of #488 + [mobility-actiontext](https://github.com/sedubois/mobility-actiontext) gem (will be available with Mobility 1.1).

sedubois:
@shioyama now that mobility-actiontext 0.1.0 has been released, should it be documented in the website/readme/wiki?

shioyama:
Sure! Could you add it to the Wiki and make a PR to add to the readme? :bowing_man:

sedubois:
- #510 adds a note to the README 
- added a one-line page to the [Wiki](https://github.com/shioyama/mobility/wiki/Integration-with-Rails-Action-Text)
- added a note at the bottom of the [KeyValue backend wiki page](https://github.com/shioyama/mobility/wiki/KeyValue-Backend#action-text)

PR #414, opened 2020-07-10
Make ActiveRecord and Sequel actual plugins

shioyama:
Currently on `master`, `Mobility::Plugins::ActiveRecord`, `Mobility::Plugins::ActiveModel` and `Mobility::Plugins::Sequel` are simply namespaces which hold modules called by `Mobility::Plugins::Dirty` depending on what gems are loaded.

This is really ugly, because it means that code that should be generic has to know every possible ORM variant.

With this PR, we use inversion of control to flip this situation around so that `active_record` and `sequel` are real plugins which themselves call `active_record_dirty` or `sequel_dirty`, `active_record_cache` or `sequel_cache`, etc. These latter plugins depend are triggered by inclusion of the generic ones (`dirty`, `cache`).

So basically, if you have:

```ruby
  plugin :active_record
  plugin :dirty
end
```

then the `active_record` plugin includes `active_record_dirty`, which itself depends on `dirty` (with `include: false`). This way, the AR specific dirty code is only triggered if you actually enable `dirty`. Same for Sequel.

The other thing that happens here is that we assume that if you have `plugin :active_record`, the model you're translating must be an `ActiveRecord::Base` model. Previously, the Dirty plugin would trigger on the `included` hook and check which class it was included into, but this meant that the model class would get more modules than necessary.

Now, everything is explicit and you must declare `plugin :active_record`, and when you do, Mobility expects translated classes to be ActiveRecord classes, etc. If (in that 0.001% of cases) you are using Mobility with both ActiveRecord _and_ Sequel, you'd just have to create two module builders, one for each:

```ruby
class ActiveRecordTranslatedAttributes < Mobility::Attributes
  plugin :active_record
  # ...
end

class SequelTranslatedAttributes < Mobility::Attributes
  plugin :sequel
  # ...
end

class ActiveRecordModel < ActiveRecord::Base
  include ActiveRecordTranslatedAttributes.new(:title)
end

class SequelModel < Sequel::Model
  include SequelTranslatedAttributes.new(:title)
end
```

So since everything is now more modular, this is still entirely possible.

The nice thing here is that all the ugly code which was implicitly trying to figure out what gems are loaded (to automatically switch between ORM classes) is gone. I think explicitness here is much clearer and safer.

Same goes for backends: the `active_record` plugin loads an `active_record_backend` plugin which overrides `load_backend` to first search for a backend under `Mobility::Backends::ActiveRecord` (same for `sequel_backend`). This again means a lot of magic becomes less magical, which I think is good.

---

Update: I've also extracted UniquenessValidation into its own plugin, which is automatically included when you call `plugin :active_record`.


lib/mobility.rb
```diff
@@ -47,44 +47,12 @@ class VersionNotSupportedError < ArgumentError; end
...
-      require "sequel/extensions/inflector"
-    end
-    require "sequel/plugins/dirty"
-    require "mobility/sequel"
-    Loaded::Sequel = true
-  rescue LoadError => e
-    raise unless e.message =~ /sequel/
-    Loaded::Sequel = false
-  end
-
```
shioyama, line 87: I will be so happy when this is all gone.

PR #418, opened 2020-07-26
Remove unnecessary const_set for uniqueness validator

shioyama:
I noticed that this const_set here is not actually doing anything, and in fact uniquenes validation works simply because namespacing leads `conts_get` to `Mobility::ActiveRecord::UniquenessValidator` before `ActiveRecord::Validations::UniquenessValidator`.

This is probably not good... we'd prefer to have a new subclass then to magically resolve to the Mobility validator class. Need to figure out a better way.

### DIFF
--- a/lib/mobility/active_record.rb
+++ b/lib/mobility/active_record.rb
@@ -11,13 +11,7 @@ Module loading ActiveRecord-specific classes for Mobility models.
     require "mobility/active_record/uniqueness_validator"
 
     def self.included(model_class)
-      model_class.class_eval do
-        unless const_defined?(:UniquenessValidator)
-          const_set(:UniquenessValidator,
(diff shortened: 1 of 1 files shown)


shioyama:
Related to #258. The way uniqueness validation is currently working depends on exactly the namespacing that is causing the problem in that issue. I'll try to fix this so tihngs don't work this way in v1.0.

lib/mobility/active_record.rb
```diff
@@ -11,13 +11,7 @@ module ActiveRecord
     require "mobility/active_record/uniqueness_validator"
 
     def self.included(model_class)
-      model_class.class_eval do
-        unless const_defined?(:UniquenessValidator)
```
shioyama, line 15: Figured out the problem: this should be `const_defined?(:UniquenessValidator, false)`.

Issue #411, opened 2020-07-08
Initializer test is missing


sedubois:
There is currently no test verifying that the initializer (e.g as it is generated by `rails g mobility:install` is correct.

Ref: https://github.com/shioyama/mobility/issues/410#issuecomment-655535732

shioyama:
Just to clarify (because I understand what you mean but it might not be clear to others), there [is](https://github.com/shioyama/mobility/blob/b24eb429b9b4ecf8cd1c4e7dae47d2e40de4d23e/spec/generators/rails/mobility/install_generator_spec.rb) an install generator spec, the problem is that it doesn't actually _run_ the generator, it just checks that the file it generates matches what the specs test for.

Currently on `master`, the spec passes but actually the install generator does not run correctly. This needs to be fixed but I'm putting it off until I finish cleaning up some other stuff so that everything can change together.

PR #423, opened 2020-08-06
Update initializer to work with v1.0 changes

shioyama:
Related: #411

### DIFF
--- a/lib/mobility/plugins/dirty.rb
+++ b/lib/mobility/plugins/dirty.rb
@@ -20,6 +20,8 @@ details.
     module Dirty
       extend Plugin
 
+      default true
+
       requires :backend, include: :before
       requires :fallthrough_accessors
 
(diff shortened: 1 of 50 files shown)


shioyama:
I realized we haven't been running generator specs for a while, because `Rails` is not defined by default in specs (we only include ActiveRecord).

lib/mobility/plugins/dirty.rb
```diff
@@ -20,6 +20,8 @@ module Plugins
     module Dirty
       extend Plugin
 
+      default true
```
shioyama, line 23: This causes a lot of problems in specs, because the specs currently run with the plugin included but disabled by default. I've explicitly set it to `false` in specs now so it can still be included, but really specs should only include this and other auto-enabled plugins when they are actually being tested; otherwise, they shouldn't be included in the first place.

PR #435, opened 2020-08-28
Clean up spec helper and fix kwargs issue with plugins block

shioyama:
This started as a simple spec cleanup but in the process I realized that the way the `plugins` block now works, it accepts optional kwargs which breaks setting a default hash in plugins that don't accept kwargs.

Mobility has to get this whole keyword arguments thing in order... but anyway the fix here is just to treat all arguments passed in the block as positional arguments, and let the backend plugin (which is the only plugin that accepts an options hash along with the backend name) handle the special case.

### DIFF
--- a/lib/mobility/plugin.rb
+++ b/lib/mobility/plugin.rb
@@ -114,7 +114,7 @@ Also includes a +configure+ class method to apply plugins to a pluggable
     # @param [Hash] defaults
     # @param [Symbol] key Plugin key on hash
     # @param [Array] args Method arguments
-    def configure_default(defaults, key, *args, **)
+    def configure_default(defaults, key, *args)
       defaults[key] = args[0] unless args.empty?
     end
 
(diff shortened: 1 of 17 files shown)


shioyama:
Ok one spec is failing and it seems to be something unrelated on Rails 6.1, not easily reproducible but for the record:

- Rails commit is 7101489293ece071013417aeed2da3aa1b0927c9
- Spec seed is 56849

I can reproduce it locally. I'm going to investigate.

shioyama:
Ok, this _seems_ to be an issue with tests, not with actual code. I find that in a spec on the table backend, I end up in a belongs to association which points to a `klass`, whose name is `Article` but where `klass == Article # => false`, so something is being left behind from other specs. The `klass` has an association from the KeyValue backend.

I suppose it would be a good idea to at least give the classes different names when using `stub_const`...

shioyama:
I've bisected this to rails/rails@62cfbdf36e79bc4d2a0556b628a55bcc403b18b4

shioyama:
This is the thing that is breaking tests:

rails/rails@62cfbdf36e79bc4d2a0556b628a55bcc403b18b4
`activerecord/lib/active_record/inheritance.rb`
```diff
       def polymorphic_class_for(name)
-        name.constantize
+        if store_full_class_name
+          ActiveSupport::Dependencies.constantize(name)
+        else
+          compute_type(name)
+        end
       end
```

shioyama:
Ok, [found it](https://github.com/shioyama/rails/blob/62cfbdf36e79bc4d2a0556b628a55bcc403b18b4/activesupport/lib/active_support/dependencies.rb#L615-L649). Something called `ClassCache`, which we'll need to `clear!`, apparently.

PR #437, opened 2020-08-28
Cleanup plugin usage in specs

shioyama:
We shouldn't be just enabling all these plugins globally in specs, since we no longer have to. Let's see if we can clean this up a bit.

Summary of changes:

## Custom spec types for Plugin/Backend

It's now more straightforward to use helpers in tests, just set `type: :backend` or `type: :plugin` and the spec will have relevant helper methods included. I think this is easier and easier to remember than a bunch of helper modules (these still exist, but they are included automatically when you assign the `type`).

## Don't call `translates` on the model, call it on the spec

A lot of tests previously called `translates` on the model class to setup backends on attributes. The problem with this is that `translates` is not the underlying mechanism for translating attributes; it just calls `include` with an instance of the module builder with the options it's been passed.

To avoid hard-coding to this method, a helper method `translates` is now included which can be called in a similar fashion:

```diff
-Article.translates :title, foo: :bar
+translates Article, :title, foo: :bar
```

This is an easy change once you get used to it.

## No models are pre-created, only tables

Previously there were files `spec/active_record/models.rb` and `spec/sequel/models.rb` with some model definitions which were included globally from the `spec_helper` whenever running any tests. I've removed those and instead we only pre-create tables for `articles`. To test against a particular model setup, use `stub_const`:

```ruby
before { stub_const 'Article', Class.new(ActiveRecord::Base) }
```

This makes each spec a clean slate which sets up its own models, including any translated attributes.

## No plugin is ever enabled globally

Now this is really the key thing I wanted from this PR. Recently I tried to change the `fallbacks` plugin to default to `true`, and a bunch of specs broke that had nothing to do with fallbacks. The problem is that plugins were being enabled globally and then "disabled" per spec by passing options to `translates` which turned them off. This is really clumsy and leaks unrelated plugin and backend settings between specs.

With the changes here, *no plugins are enabled globally across tests*. This means even very basic plugins like `Reader` and `Writer` are not enabled unless you ask for them. Conversely, this allows you to pinpoint specs on a particular aspect of behaviour without pulling in unrelated functionality.

To enable a plugin in a test or set of tests, you call `plugins` with either an array of symbols (names) or a block. Here's an example testing the AR Table backend:

```ruby
before { stub_const 'Article', Class.new(ActiveRecord::Base) }

context "with basic plugins" do
  plugins :active_record, :reader, :writer
  before { translates Article, :title, :content, backend: :table }
```

In this spec, the `active_record`, `reader` and `writer` plugins are enabled, but nothing else.


shioyama:
Slowly getting there...

spec/spec_helper.rb
```diff
@@ -21,25 +21,6 @@
...
-    dirty
-    fallbacks
-    presence
-    default
-    fallthrough_accessors
-    locale_accessors
-  end
-  config.plugin orm if orm
-  config.plugin :attribute_methods if defined?(ActiveRecord)
-end
```
shioyama, line 41: This is basically the point of this whole PR: to not set plugins globally in tests, but instead enable them selectively per test.

Issue #110, opened 2017-11-24
Mysql2::Error on Migration for mobility_string_translations when encoding is utf8mb4


eiapopeia:
## Context
I created the basic migrations with `rails generate mobility:install` with backend: `key_value`  using mobility version 0.2.3 (Rails 5.1.4).

## Expected Behavior
It should run the initial migration.

## Actual Behavior
When I do `rails db:migrate` it fails for 
```ruby
add_index(:mobility_string_translations, [:translatable_type, :key, :value, :locale], {:name=>:index_mobility_string_translations_on_query_keys})
```
with the following message:

> Mysql2::Error: Specified key was too long; max key length is 3072 bytes: CREATE  INDEX `index_mobility_string_translations_on_query_keys`  ON `mobility_string_translations` (`translatable_type`, `key`, `value`, `locale`)

## Possible Fix
This maybe happens  because we use utf8mb4-encoding in database.yml
```yml
development:
  adapter: mysql2
  encoding: utf8mb4
```
[UTF8mb4 is a 4-Byte character encoding](https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-utf8mb4.html). We prefer it because it is a real UTF-8-encoding (the origininal is missing lots of symbols, see [this explanation](https://medium.com/@adamhooper/in-mysql-never-use-utf8-use-utf8mb4-11761243e434)).
We can now calculate the length of the index:
+ `translatable_type`: 255*4
+ `key`: 255*4
+ `value`: 255*4
+ `locale`: 255*4
=============
index: **4080** Bytes
which is bigger than 3072.

We can for example make **locale** `varchar(15)` (the longest locales [I found](https://msdn.microsoft.com/en-us/library/cc233968.aspx) were 12 characters long). 

We then have to save another 12 chars. That can be for example by changing **key** from `varchar(255)` to `varchar(243)`.

So the Migration would look like this:
```ruby
class CreateStringTranslations < ActiveRecord::Migration[5.1]

  def change
    create_table :mobility_string_translations do |t|
      t.string  :locale, limit: 15 # NEW limit option
      t.string  :key, limit: 243   # NEW limit option
      t.string  :value
      t.integer :translatable_id
      t.string  :translatable_type
      t.timestamps
    end
    add_index :mobility_string_translations, [:translatable_id, :translatable_type, :locale, :key], unique: true, name: :index_mobility_string_translations_on_keys
[...]

PR #438, opened 2020-09-10
add description about insufficient length problem for mysql , mariadb

charlietag:
Both MySQL and MariaDB have the same problem.  So I add some note into the document.  Hope it's useful to remind user, that it could be happened and how to solve it through #110 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,12 @@ end
 To use a different default backend, set `default_backend` to another value (see
 possibilities [below](#backends)).
 
+If you choose `key_value` as `default_backend`, and **MySQL / MariaDB** is your
+database. Then, index-key-length problem (see
+[MySQL#innodb-limits](https://dev.mysql.com/doc/refman/8.0/en/innodb-limits.html),
+[mariadb#page-sizes](https://mariadb.com/kb/en/innodb-limitations/#page-sizes))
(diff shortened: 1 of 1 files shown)


shioyama:
Thanks! Will have a look and merge this very soon.

README.md
```diff
@@ -102,6 +102,12 @@ end
 To use a different default backend, set `default_backend` to another value (see
 possibilities [below](#backends)).
 
+If you choose `key_value` as `default_backend`, and **MySQL / MariaDB** is your
```
shioyama, line 105: The `default_backend` config has changed on the master branch (for v1.0). Could you reword this to:

> If you use `key_value` as your backend, ...

This way the meaning is more general, and I can also cherry-pick this into the 0-8-stable branch.

charlietag: @shioyama 
Thanks for your suggestion. I've pushed a new commit for this.

shioyama:
Hi @charlietag, sorry for the delay. I'm planning to rewrite the readme, so this won't fit very well right now. I'll cherry-pick it in when I figure out how things will be arranged.

charlietag:
Hi @shioyama , that's ok.
Thanks for your effort.
So I can use this so convenient gem!

PR #449, opened 2020-10-14
Remove Mobility::ClassMethods#mobility_attribute?

shioyama:
This is defined in the Attributes plugin, so doesn't need to be defined here.

### DIFF
--- a/lib/mobility.rb
+++ b/lib/mobility.rb
@@ -44,7 +44,6 @@ module Mobility
       return if model_class.respond_to? :mobility_accessor
 
       model_class.extend Translates
-      model_class.extend ClassMethods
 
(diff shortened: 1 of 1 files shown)


lib/mobility.rb
```diff
@@ -186,14 +185,6 @@ def set_locale(locale)
     end
   end
 
-  module ClassMethods
-    # Return translated attribute names on this model.
-    # @return [Array<String>] Attribute names
-    def mobility_attribute?(_)
```
shioyama, line 192: Previously this had to be defined here because it was used by the uniqueness validator, which was immediately included in a model with `include Mobility`, and would break unless there was a `translates` call. But now the uniqueness validator is only included when `translates` is called, so we don't need this.

PR #457, opened 2020-11-02
Raise an exception if invalid options are passed to Translations initializer

shioyama:
I've noticed that with the changes to how plugins and backends are now configured, it's very easy to forget to define a plugin and pass the plugin option, without realizing that the option is just being ignored. This could have very problematic consequences, so we should raise in this case.

This should be simple, but in fact because of the way that backend options are mixed in, it's somewhat ugly to implement. This works but could perhaps be improved.

### DIFF
--- a/spec/integration/active_record_compatibility_spec.rb
+++ b/spec/integration/active_record_compatibility_spec.rb
@@ -6,12 +6,12 @@ describe "ActiveRecord compatibility", orm: :active_record do
   include Helpers::Plugins
   include Helpers::Translates
   # Enable all plugins that are enabled by default pre v1.0
-  plugins :active_record, :reader, :writer, :cache, :dirty, :presence, :query, :attribute_methods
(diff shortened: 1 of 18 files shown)


spec/integration/active_record_compatibility_spec.rb
```diff
@@ -6,7 +6,7 @@
   include Helpers::Plugins
   include Helpers::Translates
   # Enable all plugins that are enabled by default pre v1.0
-  plugins :active_record, :reader, :writer, :cache, :dirty, :presence, :query, :attribute_methods
+  plugins :active_record, :reader, :writer, :cache, :dirty, :presence, :query, :attribute_methods, :fallbacks
```
shioyama, line 9: Interestingly, this PR highlighted that the spec here was not actually enabling the `fallbacks` plugin, although specs were testing fallbacks stuff. So basically those tests were doing nothing useful :facepalm:

PR #460, opened 2020-11-03
Fix Ruby 2.7.0 Deprecation Warnings

shioyama:
Replaces #357. This is an attempt to get rid of deprecation warnings on 1.0. Original PR by @f1sherman, syncing it up with changes since that PR.

### DIFF
--- a/lib/mobility/arel.rb
+++ b/lib/mobility/arel.rb
@@ -38,7 +38,7 @@ module Mobility
       attr_reader :locale
       attr_reader :attribute_name
 
-      def initialize(relation, column_name, locale, backend_class, attribute_name: nil)
+      def initialize(relation, column_name, locale, backend_class, attribute_name = nil)
         @backend_class = backend_class
         @locale = locale
         @attribute_name = attribute_name || column_name
--- a/lib/mobility/backend.rb
+++ b/lib/mobility/backend.rb
@@ -100,7 +100,7 @@ On top of this, a backend will normally:
     # @param [Symbol] locale Locale to read
     # @return [TrueClass,FalseClass] Whether translation is present for locale
     def present?(locale, options = {})
-      Util.present?(read(locale, options))
+      Util.present?(read(locale, **options))
     end
 
     # @!method model_class
--- a/lib/mobility/backends/active_record/hstore.rb
+++ b/lib/mobility/backends/active_record/hstore.rb
@@ -18,7 +18,7 @@ Implements the {Mobility::Backends::Hstore} backend for ActiveRecord models.
 
         # @!macro backend_writer
         def write(locale, value, options = {})
-          super(locale, value && value.to_s, options)
+          super(locale, value && value.to_s, **options)
         end
         # @!endgroup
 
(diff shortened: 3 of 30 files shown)


shioyama:
Alright, looks like different Ruby versions do things differently, this will take a bit more work...

shioyama:
Realized we weren't running the perf specs :facepalm: I fixed it so we now run them, and with this change the accessors are still within test limits, but the `Mobility::Translations` initializer test allocation count more than doubles, from a limit of 25 to 54 objects created. Seems like it's from all the plugin separation, but I think that's a very worthwhile tradeoff. The more important one is calling accessors and those numbers don't change, so I think we're fine.

shioyama:
Oh yeah, it's because we now have separated the reader/writer/backend modules and each one overrides `initialize`, and each time that is called we generate a hash due to `**options`. So it is partly related to the kwargs change. I think it's fine, hopefully Ruby will make using kwargs like that not create unnecessary hashes soon.

f1sherman:
Thank you, can't wait to get this in our codebase!

shioyama:
@f1sherman Let me know if you have issues upgrading, I'd be eager to hear. Need confirmation of successful upgrades before I can make a (non-pre-release) 1.0 release.

f1sherman:
@shioyama sounds great, I'll aim to give it a whirl this week. Thanks!

f1sherman:
@shioyama any chance you could publish an updated version of [mobility-ransack](https://rubygems.org/gems/mobility-ransack) that is compatible with mobility 1.x?

shioyama:
It's on my list of things to do!

f1sherman:
@shioyama I just integrated mobility 1.0.0.beta1 and mobility-ransack 1.0.0.alpha and all of our unit and integration tests are passing. I'm also not seeing any more deprecation warnings in our logs. I'll plan to do some manual smoke testing as well.

One note: we were using the `translated_attribute_names` method in a few places where we enumerate the attributes that have translations. I swapped those to use the protected `mobility_attributes` method instead, which seems to be working fine but isn't ideal. Any advice on what to do there?

Thank you!

shioyama:
Great to hear everything's working! 

> I'll plan to do some manual smoke testing as well.

Definitely that would be great. Mostly the change in 1.0 is a reorganization which is visible mostly in config and how plugins are written (so Mobility can be more easily extensible), but stuff should work the same otherwise.

> One note: we were using the translated_attribute_names method in a few places where we enumerate the attributes that have translations. I swapped those to use the protected mobility_attributes method instead, which seems to be working fine but isn't ideal. Any advice on what to do there?

You can get that by adding the `attribute_methods` plugin to your config. It's just not default included anymore.

shioyama:
I should mention though that I might be changing that plugin before releasing 1.0. I'll update if I do though.

f1sherman:
Sounds great, thanks!

f1sherman:
@shioyama just wanted to give you a heads up that my smoke testing went well. I'll let you know if I run into any issues but for now everything looks good!

f1sherman:
@shioyama I noticed the following note for the `attribute_methods` plugin:
```
@note Adding translated attributes to +attributes+ can have unexpected
  consequences, since these attributes do not have corresponding columns in the
  model table. Using this plugin may lead to conflicts with other gems.
```

If possible, it would be nice to just get the `translated_attribute_names` method (or something equivalent) without needing to incur the risk noted above. Not a major issue, just wanted to pass that along. Thanks!

shioyama:
Yes, that's a very good point. I will find a less invasive way to make mobility attributes available, maybe I'll just change the method from protected back to public.

PR #462, opened 2020-11-08
Refactor attributes & backend plugins

shioyama:
Summary:
- make `mobility_attributes` public (see https://github.com/shioyama/mobility/pull/460#issuecomment-722445031)
- remove `register_mobility_attribute`, which was public but intended to be private (actually don't need it)
- add tests

Also, while I was add it I made a change to the backend plugin to remove `register_mobility_backend_class`, which we don't actually need either.

### DIFF
--- a/lib/mobility/plugins/attributes.rb
+++ b/lib/mobility/plugins/attributes.rb
@@ -55,8 +55,6 @@ for aggregating attributes.
           mobility_attributes.each { |name| klass.register_mobility_attribute(name) }
         end
 
-        protected
-
         # Return translated attribute names on this model.
         # @return [Array<String>] Attribute names
         def mobility_attributes
--- a/lib/mobility/plugins/attributes.rb
+++ b/lib/mobility/plugins/attributes.rb
@@ -32,8 +32,15 @@ for aggregating attributes.
       end
 
       included_hook do |klass|
-        klass.extend ClassMethods
-        @names.each { |name| klass.register_mobility_attribute(name) }
+        names = @names
+
+        klass.class_eval do
+          extend ClassMethods
+          names.each { |name| mobility_attributes << name.to_s }
+          mobility_attributes.uniq!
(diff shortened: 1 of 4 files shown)


shioyama:
@f1sherman this should do the trick :wink:

f1sherman (approved):
Amazing, thank you so much!

lib/mobility/plugins/attributes.rb
```diff
@@ -44,24 +48,17 @@ def mobility_attribute?(name)
...
         # Return translated attribute names on this model.
         # @return [Array<String>] Attribute names
         def mobility_attributes
           @mobility_attributes ||= []
         end
+
+        def inherited(klass)
+          super
+          attrs = mobility_attributes
+          klass.class_eval { @mobility_attributes = attrs }
```
shioyama, line 60: This is not right, we should be duping `attrs` otherwise both classes are pointing at the same object.

shioyama:
So there is a rare situation where things can go weird, looks like this:

```ruby
class Foo
  extend Mobility
  translates :title
end

class Bar < Foo
end

Foo.translates :content

Foo.mobility_attributes
#=> ["title", "content"]

Bar.mobility_attributes
#=> ["title"]
```

`Bar` should have the same attributes as `Foo`, since it inherits all of them, but it only has `"content"`.

The problem is that after its attributes being inherited in `Bar`, further attributes added to `Foo` will not be updated in `Bar`. While this situation could be handled, it requires trickier code and I think this situation is very rare.

That said, I don't want to allow this to happen, so what I've done is to `freeze` the attribute of the parent after being subclassed. This way, if you try to actually do this, you'll get:

```ruby
Mobility::Plugins::Attributes::FrozenAttributesError: Attempting to translate these attributes on Foo, which has already been subclassed: content.
```

Not sure this is the most obvious message, but I think it'll do until someone is confused by it :sweat_smile:

PR #464, opened 2020-11-15
Last changes before release of 1.0.0.rc1

### LINKED ISSUE #258: NameError (uninitialized constant Mobility::ActiveRecord::VERSION)

shioyama:
Just doing some cleanup, we'll see how this goes.

Among other things, these changes entirely remove the `Mobility::ActiveRecord` namespace, which fixes #258. :tada:

- [X] fix keyword argument deprecation errors using Sequel
- [X] add inline documentation for two ways to translate attributes
- [X] remove `Mobility::ActiveRecord` namespace (fixes #258) 
- [x] remove `Mobility::Arel` namespace
- [x] remove `Mobility::Sequel` namespace


shioyama:
Things left to do:
- remove `Mobility::Sequel` namespace
- remove `Mobility::Arel` namespace

PR #468, opened 2020-12-06
Fix setting a locale from Rails config

sergey-alekseev:


lib/mobility.rb
```diff
@@ -218,7 +218,7 @@ def enforce_available_locales!(locale)
     #   methods (in LocaleAccessors) than is really necessary.
     def available_locales
       if defined?(Rails) && Rails.application
-        Rails.application.config.i18n.available_locales || I18n.available_locales
+        Rails.application.config.i18n.available_locales&.map(&:to_sym) || I18n.available_locales
```
shioyama, line 221: Would we expect these to be strings? I don't really have any experience setting available locales in Rails but it seems most examples (e.g. [here](https://lokalise.com/blog/rails-i18n/) and [here](https://phrase.com/blog/posts/rails-i18n-setting-and-managing-locales/)) have them as symbols.

I'm ok with this change but just curious about the context for it.

shioyama (approved):

shioyama:
Merged in d8c2e944accc0ed115eb9aca82ad97faac7cbf56

shioyama:
Thanks for the great tests BTW :bow: Always appreciate that!

lib/mobility.rb
sergey-alekseev: Indeed, probably majority of usage examples are symbols. However strings are allowed as well and for sure can be used.  

As for the context – good question. I added this because we set `config.i18n.available_locales` to a value from settings which uses [the config gem](https://github.com/rubyconfig/config). The gem doesn't support symbols as values parsed from YAML. So the value is an array of strings and that's why I encountered this problem and submitted a patch.

shioyama: Thanks for the context! That makes sense.

PR #470, opened 2020-12-10
Fixes validations with validates_uniqueness_of

artplan1:
Hi, we're having issues with uniqueness validations done with `validates_uniqueness_of`.
I assume that constant in `validates_uniqueness_of ` is still points to default rails validator.

```ruby
    Failure/Error: model_class.create(attribute1 => "foo")

     ActiveRecord::StatementInvalid:
       PG::UndefinedColumn: ERROR:  column container_posts.title does not exist
       LINE 1: SELECT 1 AS one FROM "container_posts" WHERE "container_post...
```

I came up with straightforward solution, but I believe there are better solutions.

### DIFF
--- a/lib/mobility/plugins/active_record/uniqueness_validation.rb
+++ b/lib/mobility/plugins/active_record/uniqueness_validation.rb
@@ -10,6 +10,10 @@ module Mobility
           klass.class_eval do
             unless const_defined?(:UniquenessValidator, false)
               self.const_set(:UniquenessValidator, Class.new(UniquenessValidator))
+
+              def self.validates_uniqueness_of(*attr_names)
+                validates_with(UniquenessValidator, _merge_attributes(attr_names))
+              end
             end
           end
         end
--- a/spec/support/shared_examples/validation_examples.rb
+++ b/spec/support/shared_examples/validation_examples.rb
@@ -26,6 +26,19 @@ shared_examples_for "AR Model validation" do |model_class_name, attribute1=:titl
       end
     end
 
+    context "with validates_uniqueness_of" do
+      let(:model_class) do
+        model_class = model_class_name.constantize
+        model_class.validates_uniqueness_of(attribute1)
+        model_class
+      end
+
+      it "is invalid if other record has same attribute value in this locale" do
+        model_class.create(attribute1 => "foo")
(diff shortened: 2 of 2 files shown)


shioyama:
Thanks for the PR! :heart: 

> I came up with straightforward solution, but I believe there are better solutions.

Hmm well I think it's fairly safe to do it this way, since this only overrides the method on the class and not other classes.

For reference, the reason the code has `const_get`/`const_set` is because `validates` looks for the constant `UniquenessValidator` and will prefer the model namespace if there is a constant defined there:

https://github.com/shioyama/rails/blob/5262ec59002e461b2fdb8cb81f0ac1623b1cd2fd/activemodel/lib/active_model/validations/validates.rb#L119
```ruby
            validator = key.include?("::") ? key.constantize : const_get(key)
```

So by defining the constant, we can selectively override the default uniqueness validator on this model only. If that wasn't possible we'd have to override `validates` which would be much more tricky.

But `validates_uniqueness_of` can easily be overridden on the model since it's just a one-line method. So I think this is a :+1: solution.

spec/support/shared_examples/validation_examples.rb
```diff
@@ -26,6 +26,19 @@
...
+        model_class = model_class_name.constantize
+        model_class.validates_uniqueness_of(attribute1)
+        model_class
+      end
+
+      it "is invalid if other record has same attribute value in this locale" do
+        model_class.create(attribute1 => "foo")
+        expect(model_class.new(attribute1 => "foo")).not_to be_valid
+      end
+    end
```
shioyama, line 40: Thanks for the test!

lib/mobility/plugins/active_record/uniqueness_validation.rb
```diff
@@ -10,6 +10,10 @@ module UniquenessValidation
           klass.class_eval do
             unless const_defined?(:UniquenessValidator, false)
               self.const_set(:UniquenessValidator, Class.new(UniquenessValidator))
+
+              def self.validates_uniqueness_of(*attr_names)
+                validates_with(UniquenessValidator, _merge_attributes(attr_names))
+              end
```
shioyama, line 16: LGTM :+1:

shioyama (approved):

PR #471, opened 2020-12-12
Make Mobility::Plugins::ActiveRecord::Query::VirtualRow public

shioyama:
There should be a way to build queries without using the query scope. This PR makes this more easily possible by making `Mobility::Plugins::ActiveRecord::Query::VirtualRow` public, and also adding a class method `Query.build_query` which can be used like this:

```ruby
Mobility::Plugins::ActiveRecord::Query.build_query(Post) do
  title.eq('foo')
end
```

This is equivalent to:
```ruby
Post.i18n do
  title.eq('foo')
end
```

The namespace is cumbersome but you can just name it something else in your application:

```ruby
I18nQueryScope = Mobility::Plugins::ActiveRecord::Query
```

then use:

```ruby
I18nQueryScope.build_query(Post) do
  title.eq('foo')
end
```

### DIFF
--- a/lib/mobility/plugins/active_record/query.rb
+++ b/lib/mobility/plugins/active_record/query.rb
@@ -29,7 +29,9 @@ enabled for any one attribute on the model.
             klass.class_eval do
               extend QueryMethod
               extend FindByMethods.new(*plugin.names)
-              singleton_class.send :alias_method, plugin.query_method, :__mobility_query_scope__
+              singleton_class.define_method(plugin.query_method) do |locale: Mobility.locale, &block|
+                Query.build_query(self, locale, &block)
+              end
             end
             backend_class.include BackendMethods
           end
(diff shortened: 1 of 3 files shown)


sedubois:
> There should be a way to build queries without using the query scope

Thank you @shioyama, what does `Mobility::Plugins::ActiveRecord::Query.build_query(Post)` improve compared to `Post.i18n`?

shioyama:
> what does `Mobility::Plugins::ActiveRecord::Query.build_query(Post)` improve compared to `Post.i18n`?

Not a whole lot, but it takes away the requirement to define a class method (`i18n`) on the model. Internally, code can avoid depending on that method and instead reference `Mobility::Plugins::ActiveRecord::Query.build_query`.

Since the method name is configurable other plugins and gems shouldn't rely on the method name, so this would provide a "safe" way to get the same functionality.

PR #473, opened 2020-12-12
Use cache v2 in CI

artplan1:
v2 increases speed of caching and reduces size
https://github.com/actions/cache/releases


shioyama:
Great, the faster the better :+1: Thanks! :heart:

PR #479, opened 2020-12-19
Allow setting locale per node in query block

shioyama:
I realized that querying using the block format can be made much more powerful if we allow passing the locale as an argument to the node method in the block, e.g.:

```ruby
Post.i18n do
  title(:en).eq('foo').or(content(:ja).eq('bar'))
end
```

So the argument to `title` and `content` in the block is optional, but if passed will override the locale and apply it just to that node. With this change, it becomes really easy to do very complex queries across multiple languages in the same block.

WDYT folks?

### DIFF
--- a/spec/mobility/plugins/sequel/query_spec.rb
+++ b/spec/mobility/plugins/sequel/query_spec.rb
@@ -1,0 +1,160 @@
+require "spec_helper"
+
+return unless defined?(Sequel)
+
+require "mobility/plugins/sequel/query"
+
+describe Mobility::Plugins::Sequel::Query, orm: :sequel, type: :plugin do
+  plugins :sequel, :reader, :writer, :query, :cache
+  plugin_setup
+
+  describe "query_scope" do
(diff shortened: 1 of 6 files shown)


shioyama:
Added Sequel support as well.

spec/mobility/plugins/sequel/query_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+    it "creates a query method method" do
+      stub_const 'Article', Class.new(Sequel::Model)
+      Article.dataset = DB[:articles]
+      translates Article, :title, backend: :table
+      article = Article.create(title: "foo")
+      expect(Article.i18n.first).to eq(article)
+    end
+  end
+
+  describe "virtual row handling", focus:true do
```
shioyama, line 63: :scream:

sedubois: 😄

PR #480, opened 2020-12-20
Include query_scope in find_by methods

shioyama:
It makes sense to include the query scope automatically in the find_by methods (`find_by_title` etc) so they work out of the box.

Assume `title` is a translated attribute:

_Before_

```ruby
Post.i18n.find_by_title('foo')
#=> <Post ...>
Post.find_by_title('foo')
#=> exception
```

_After_

```ruby
Post.i18n.find_by_title('foo')
#=> <Post ...>
Post.find_by_title('foo')
#=> <Post ...>
```


madhums:
Would be nice to merge this! @shioyama

PR #487, opened 2020-12-30
Call after_destroy in setup block to delete associated translations

shioyama:
Currently we include a module which adds an `after_destroy` hook to destroy both text and string translations after a model is destroyed, if it has any attributes using the KeyValue backend. This is a conservative move to ensure that string/text translations for a model are always cleaned up, even if a model may have changed code in the past from string to text (leaving behind orphan translations, for example).

However, this makes assumptions about the `translations_class`, which in general may use another class/table entirely.

~I think dealing with orphan translations should be the responsibility of the application developer, and the code itself should only define callbacks for _currently defined attributes_. Thus if a model only has string translations, then the `after_destroy` block should only destroy string translations for that model table, not text translations. This avoids useless queries, at the cost of possibly leaving behind previously generated translations, which become orphaned.~

I've rethought this and decided to keep the current behaviour the default, with one small change: we now look for descendants of `Mobility::Backends::ActiveRecord::KeyValue::Translation`, so if someone wants to add a subclass for a type other than `:string` or `:text`, that would be possible and the `after_destroy` callback would automatically clean those up too.

I think given how easy it is to miss orphan translations, the default should be to destroy all string/translation/whatever translations for a model when it is destroyed, regardless of the current use of translations in the model. This ensures that _as long as the model is still translated with the KeyValue backend_, it will continue to cleanup all translations when records are destroyed.

### DIFF
--- a/lib/mobility/backends/active_record/key_value.rb
+++ b/lib/mobility/backends/active_record/key_value.rb
@@ -150,8 +150,8 @@ Implements the {Mobility::Backends::KeyValue} backend for ActiveRecord models.
       end
 
       setup do |attributes, options|
-        association_name   = options[:association_name]
-        translations_class = options[:class_name]
+        association_name  = options[:association_name]
+        translation_class = options[:class_name]
 
         # Track all attributes for this association, so that we can limit the scope
         # of keys for the association to only these attributes. We need to track the
@@ -164,7 +164,7 @@ Implements the {Mobility::Backends::KeyValue} backend for ActiveRecord models.
 
         has_many association_name, ->{ where key: association_attributes },
           as: :translatable,
-          class_name: translations_class.name,
+          class_name: translation_class.name,
           inverse_of: :translatable,
           autosave:   true
         before_save do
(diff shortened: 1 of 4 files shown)


lib/mobility/backends/active_record/key_value.rb
```diff
@@ -186,7 +186,12 @@ def visit_default(_)
           include const_set(module_name, callback_methods)
         end
 
-        include DestroyKeyValueTranslations
+        after_destroy do
+          klasses = [translations_class, *Mobility::Backends::ActiveRecord::KeyValue::Translation.descendants].uniq
```
shioyama, line 190: I reached a kind of compromise: we destroy any descendant of `Mobility::Backends::ActiveRecord::KeyValueTranslation`, plus `translations_class` if that class is defined and different. This may be more than is necessary but I think it's safest to ensure all translations are deleted.

lib/mobility/backends/active_record/key_value.rb
```diff
@@ -198,17 +203,6 @@ def translation_for(locale, **)
         translation
       end
 
-      module DestroyKeyValueTranslations
```
shioyama, line 201: I realized that there's really no point using a module for this since including it multiple times will continue to trigger the `included` hook (the whole point was to only trigger the hook the first time, but that's not how Ruby works).

shioyama:
Related: https://github.com/shioyama/mobility/pull/412

PR #488, opened 2020-12-30
Add key, value and belongs_to options to key-value backend

sedubois:
In preparation for #385.

Also see #412.

### DIFF
--- a/lib/mobility/backends/active_record/key_value.rb
+++ b/lib/mobility/backends/active_record/key_value.rb
@@ -48,7 +48,7 @@ Implements the {Mobility::Backends::KeyValue} backend for ActiveRecord models.
         #   translation table value column
         def build_node(attr, locale)
           aliased_table = class_name.arel_table.alias(table_alias(attr, locale))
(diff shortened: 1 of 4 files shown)


shioyama:
Hey wow great work! And fast!!

This looks good, just wondering about the name `translatable`, since it's not really clear what that means without reading the code. Maybe something like `belongs_to_association`?

sedubois:
@shioyama OK translatable has been renamed to belongs_to.

shioyama:
Thanks!! This will go out in v1.1 (soonish).

sedubois:
@shioyama I overlooked these lines in the PR, how to make them use the new options?

https://github.com/shioyama/mobility/blob/b187ad45b5c75c1524c56fd69bdbc222b619e76f/lib/mobility/backends/active_record/key_value.rb#L217-L220

shioyama:
I think it's fine not to go that far. The new columns should be understood to be meant for the case when a different (custom) translation class is used.

shioyama:
This has been released in [1.1.0](https://rubygems.org/gems/mobility/versions/1.1.0) :tada:

PR #489, opened 2021-01-10
Add key, value and belongs_to options to Sequel key-value backend

shioyama:
Port #488 to Sequel. Also, prefer `__send__` over `send`.

This is mostly just for completeness, so the code is consistent.

### DIFF
--- a/spec/mobility/backends/sequel/key_value_spec.rb
+++ b/spec/mobility/backends/sequel/key_value_spec.rb
@@ -146,6 +146,42 @@ describe "Mobility::Backends::Sequel::KeyValue", orm: :sequel, type: :backend do
       end
     end
 
+    context "with custom translation class" do
+      before(:all) do
+        DB.create_table!(:mobility_custom_translations) do
+          primary_key :id
+          String      :locale,      allow_null: false
+          String      :my_key
+          String      :my_value
+          Integer     :custom_id,   allow_null: false
+          String      :custom_type, allow_null: false
+          DateTime    :created_at,  allow_null: false
+          DateTime    :updated_at,  allow_null: false
+        end
+      end
+      after(:all) { DB.drop_table?(:mobility_custom_translations) }
+
+      before do
+        stub_const('CustomTranslation', Class.new(::Sequel::Model(:mobility_custom_translations)))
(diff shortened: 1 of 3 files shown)


shioyama:
This is in fact very difficult to test without making `Mobility::Backends::Sequel::KeyValue::Translation` customizable, but this would greatly complicate implementation... 

For now I'll add a test to the AR backend implementation which is easier since ActiveRecord supports polymorphic associations (whereas Sequel does not, out of the box).

shioyama:
Follow up is to use `__send__` everywhere where we are calling `send` on models outside of Mobility, since we should really not assume that `send` has not been defined for something else.

shioyama:
cc @sedubois

shioyama:
Ok, after a bit of work I got a test for Sequel KeyValue with the custom translation class and different columns. In the process, I realized that the new AR options in #488 aren't actually working because we didn't add the `key_column`, `value_column` and `belongs_to` options to `valid_keys`. I made a change here to do that.

spec/mobility/backends/sequel/key_value_spec.rb
```diff
@@ -146,6 +146,39 @@
...
+        CustomTranslation.def_column_alias :translatable_type, :custom_type
+        translates Article, :title, :content, backend: :key_value, key_column: :my_key, value_column: :my_value, belongs_to: :custom, class_name: CustomTranslation, association_name: :custom_translations
+      end
+
+      it "correctly stores translations on custom table" do
+        expect {
+           Article.create(title: "foo title")
+        }.to change { CustomTranslation.count }.by(1)
+
+      end
```
shioyama, line 179: TODO: add more here!

sedubois:
@shioyama not exactly sure what you mean by “are not actually working” as I was already able to use the custom backend on top of Mobility master.

shioyama:
> not exactly sure what you mean by “are not actually working” as I was already able to use the custom ActionText backend on top of Mobility master.

Yes because these values are set (as defaults) _inside_ the backend. But if you try to pass the new options (`key_column`, `value_column` etc) to `translates` you will get an exception because they are not included in `valid_keys`.

sedubois:
Gotcha

PR #495, opened 2021-01-22
Correctly override default backend options

### LINKED ISSUE #492: InvalidOptionKey error when using different backends
Hi Chris,

I ran into an Mobility::Plugins::Backend::InvalidOptionKey error when I use container backend

## Context
In my mobility.rb initializer

`backend :jsonb, column_suffix: '_translations'`

In my content.rb model

```
class Content < ApplicationRecord
  # t.jsonb :custom_fields
  Mobility.available_locales.each { |locale| store_accessor :custom_fields, "vimeo_url_#{locale}" }

  # Translates
  extend Mobility
  translates :vimeo_url, backend: :container, column_name: 'custom_fields'
end
```

## Expected Behavior
I expect to use `Content.new(vimeo_url: 'xxxxx')` or `cnt_instance.vimeo_url = 'xxxxx'`

## Actual Behavior
I get this error Mobility::Plugins::Backend::InvalidOptionKey (These are not valid jsonb backend keys: column_suffix.)
It seems to mix the default backend options and the specific one

Best regards,
Kevyn

shioyama:
Fixes #492

This is not a super clean fix, but tests check that it's doing what one would expect.

cc @kevynlebouille

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,11 +4,13 @@ on:
   push:
     branches:
       - master
-      - 0-8-stable
+      - 1-0
+      - 0-8
   pull_request:
     branches:
       - master
-      - 0-8-stable
+      - 1-0
+      - 0-8
 
 jobs:
   test:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@
 v1.0](https://github.com/shioyama/mobility/wiki/Introduction-to-Mobility-v1.0)
 for more details on how to upgrade.
 
+### 1.0.4
+- Correctly override default backend options, fixes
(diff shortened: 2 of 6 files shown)


shioyama:
After thinking this over for a bit, I think I'm going to use this as an opportunity to cleanly separate "backend options" from "plugin options". We'll still support passing them together in the options hash to `translates` but internally they will always be handled separately. I think in the long term this will be an improvement.

shioyama:
Ok for now I think this will do. If this passes I'll merge later and release in 1.0.4.

shioyama:
Released as [1.0.4](https://rubygems.org/gems/mobility/versions/1.0.4). @kevynlebouille can you check the new version works?

PR #505, opened 2021-02-05
Make models work with Marshal.dump

### LINKED ISSUE #494: Cannot Marshal.dump
Will it be possible to Marshal.dump a AR that extend Mobility

## Context

config/initializers/mobility.rb
```ruby
Mobility.configure do 
  plugins do
    backend :table
    active_record
    reader
    writer
    backend_reader
    query
  end  
end
```

app/models/project.rb
```ruby
class Project < ActiveRecord::Base
  extend Mobility
  translates :title
end
```

## Actual Behavior

```ruby
p = Project.new(title: "My title")
Marshal.dump(p)
...
TypeError (can't dump hash with default proc)
```

shioyama:
Fixes #494

Basically, gives names to some stuff that was anonymous.

### DIFF
--- a/lib/mobility/backend.rb
+++ b/lib/mobility/backend.rb
@@ -1,4 +1,6 @@
 # frozen-string-literal: true
+require 'digest/md5'
+
 module Mobility
 =begin
 
@@ -123,7 +125,10 @@ On top of this, a backend will normally:
     # Extend included class with +setup+ method and other class methods
     def self.included(base)
       base.extend ClassMethods
-      base.singleton_class.attr_reader :options, :model_class
+      base.singleton_class.class_eval do
+        attr_accessor :options, :model_class
+        protected :options=, :model_class=
+      end
     end
 
     # Defines setup hooks for backend to customize model class.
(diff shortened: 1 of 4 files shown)


shioyama:
This looks like it works. Note that it uses an MD5 hash of the sorted backend options in the backend subclass name, so if a code change changes these options, then Marshal will fail to find the class.

This is a pretty typical problem with Marshal since it dumps everything, so the usual caveats: don't store Marshal-serialized objects in the db and if you're dumping them to `Rails.cache` be careful to clear the cache on deploys that might change cached values (in this case if you change options to `translates`, you'll need to clear any caches that have serialized copies of the objects).

lib/mobility/backend.rb
```diff
@@ -1,4 +1,6 @@
 # frozen-string-literal: true
+require 'digest/md5'
```
shioyama, line 2: I'm a bit unsure whether this is ok. Any application that uses Mobility will now require md5. I just want a short unique hash for the options hash... :thinking:

maxime-carbonneau: I would say it is not a big deal since this is part of Ruby Standard Library.

maxime-carbonneau:
@shioyama Do you have a schedule to release this modification?

shioyama:
@maxime-carbonneau Hey sorry I haven't forgotten about this. It works but I'm not so happy about the md5 thing, I'd like to try to avoid having that in the serialization. I have a vague idea of how to do it differently but just need a bit of time to try.

maxime-carbonneau:
@shioyama I may have miss something, but why don't you use rand or a unique counter instead.

shioyama:
Random stuff or counters won't work because the same configured backend class needs to always return the same name - it has to be deterministic, otherwise when you load it back it won't match and it will fail.

maxime-carbonneau:
Could it be solved by a pseudo md5 : 
<code>
(options.sort.to_s.bytes.each_with_index.sum {|a, b| a + b } % 1_000_000).to_s
</code>
It won't be as "secure" as md5 but it may resolve the issue.

shioyama:
It's not the security that concerns me, it's the arbitrariness of it. I'd like to see if we can avoid assigning a name to the class at all, but I'm not sure that's possible. Give me a couple days to try.

shioyama:
Replaced by #532.

Issue #335, opened 2019-08-16
Option to use original column for default locale


wvengen:
Proposal: add an option to use the original column for the default locale, instead of putting the default locale in backend storage.

## Context
When adding Mobility to an existing application (with no other model translation infrastructure in place), there are usually existing columns that need translations. As far as I could see, Mobility expects all translations to live in backend storage, meaning that the original column is abandoned (or perhaps used once in a migration, after which is it abandoned - #315).

When there are many references to the original column, or if there are other applications using the same database (like SQL queries in an external tool), they all need to be rewritten. This can be a large barrier to adoption, and make it harder to change the used Mobility backend later on.

## Suggested solution
Add an option to allow the use of the original column for the default locale.

shioyama:
You should be able to achieve this with:

```ruby
class Post < Application Record
  def title(**opts)
    if I18n.locale == I18n.default_locale
      super(**opts, super: true)
    else
      super
    end  
end
```

I haven't tried the code above, but basically the idea is that `super` is a Mobility option which allows you to call through to whatever the original method was (defined by ActiveRecord for the `title` column in this case). So when you call `post.title(super: true)` you get the actual value of the `title` column. But you only want to do that when the locale is the default locale, hence the conditional.

The code above may require some tweaking.

shioyama:
I agree though that this is a common request and that there should eventually be an option or plugin for it.

wvengen:
Thanks for your quick response! And for the code example. One could indeed write a concern that adapts translated attributes like that (and the setter, and perhaps some others).
Great to hear this suggestion is welcome.

shioyama:
Glad that helped!

Just added the setter code. Also realized that this would not work with querying, and that one would be trickier to get working.

Anyway I'll leave this open for now as a feature, although I doubt I'll be getting to it for a while.

yourtallness:
Ran into a similar issue when I added mobility to my app today.

All product names became empty because there are no translation records yet and the gem does not use the original column.

For existing apps are we supposed to create a data migration to create translation records for the source language?

shioyama:
> For existing apps are we supposed to create a data migration to create translation records for the source language?

Yes, there is basically no support for doing this at the moment, although it's something I've thought about now and again. It should eventually be supported, but it's just never gotten the priority to make it happen. I don't really see that changing any time soon unless somebody from the community wants to take it on.

mattboldt:
[...]

PR #512, opened 2021-04-09
ColumnFallback plugin to fallback to original column

shioyama:
Solves: #335

I wanted to see if this would work, and it does! This works like #335 but supports any backend. (Only lightly tested mind you.)

@mattboldt if you have time, have a look. I think with some tests and a bit of effort to work out the edge cases, this could be a really great new plugin. A lot of people want this.

### DIFF
--- a/lib/mobility/plugins/active_record.rb
+++ b/lib/mobility/plugins/active_record.rb
@@ -4,6 +4,7 @@ require_relative "./active_record/dirty"
 require_relative "./active_record/cache"
 require_relative "./active_record/query"
 require_relative "./active_record/uniqueness_validation"
+require_relative "./active_record/column_fallback"
 
 module Mobility
 =begin
@@ -24,6 +25,7 @@ dirty for active_record_dirty) is also enabled.
       requires :active_record_cache
       requires :active_record_query
       requires :active_record_uniqueness_validation
+      requires :active_record_column_fallback
 
 
       included_hook do |klass|
--- a/lib/mobility/plugins/active_record/column_fallback.rb
+++ b/lib/mobility/plugins/active_record/column_fallback.rb
@@ -1,0 +1,139 @@
+# frozen-string-literal: true
+
+module Mobility
+  module Plugins
+=begin
+
+Plugin to use an original column for a given locale, and otherwise use the backend.
+
+=end
+    module ActiveRecord
+      module ColumnFallback
+        extend Plugin
+
+        requires :column_fallback, include: false
+
+        included_hook do |_, backend_class|
+          case (column_fallback = options[:column_fallback])
+          when TrueClass
+            backend_class.include I18nDefaultLocaleBackend
+          when Array, Proc
+            backend_class.include BackendModule.new(column_fallback)
+          else
+            raise ArgumentError, "column_fallback value must be a boolean, an array of locales or a proc"
+          end
+        end
+
+        module I18nDefaultLocaleBackend
+          def read(locale, **)
+            locale == I18n.default_locale ? model.read_attribute(attribute) : super
+          end
+
+          def write(locale, value, **)
+            locale == I18n.default_locale ? model.send(:write_attribute, attribute, value) : super
+          end
+
+          def self.included(base)
+            base.extend(ClassMethods)
+          end
+
+          module ClassMethods
+            def build_node(attr, locale)
+              if locale == I18n.default_locale
+                model_class.arel_table[attr]
+              else
+                super
+              end
+            end
+          end
+        end
+
+        class BackendModule < Module
+          def initialize(column_fallback)
+            case (@column_fallback = column_fallback)
+            when Array
+              define_array_accessors
+            when Proc
+              define_proc_accessors
+            end
+          end
+
+          def included(base)
+            base.extend(ClassMethods.new(@column_fallback))
+          end
+
+          private
+
+          def define_array_accessors
+            column_fallback = @column_fallback
+
+            module_eval <<-EOM, __FILE__, __LINE__ + 1
+            def read(locale, **)
+              if #{column_fallback}.include?(locale)
+                model.read_attribute(attribute)
+              else
+                super
+              end
+            end
+
+            def write(locale, value, **)
+              if #{column_fallback}.include?(locale)
+                model.send(:write_attribute, attribute, value)
+              else
+                super
+              end
+            end
+            EOM
+          end
+
+          def define_proc_accessors
+            column_fallback = @column_fallback
+
+            define_method :read do |locale, **options|
+              if column_fallback.call(locale)
+                model.read_attribute(attribute)
+              else
+                super(locale, **options)
+              end
+            end
+
+            define_method :write do |locale, value, **options|
+              if column_fallback.call(locale)
+                model.send(:write_attribute, attribute, value)
+              else
+                super(locale, value, **options)
+              end
+            end
+          end
+
+          class ClassMethods < Module
+            def initialize(column_fallback)
+              case column_fallback
+              when Array
+                module_eval <<-EOM, __FILE__, __LINE__ + 1
+                def build_node(attr, locale)
+                  if #{column_fallback}.include?(locale)
+                    model_class.arel_table[attr]
+                  else
+                    super
+                  end
+                end
+                EOM
+              when Proc
+                define_method(:build_node) do |attr, locale|
+                  if column_fallback.call(locale)
+                    model_class.arel_table[attr]
+                  else
+                    super(attr, locale)
+                  end
+                end
+              end
+            end
+          end
+        end
+      end
+    end
+
+    register_plugin(:active_record_column_fallback, ActiveRecord::ColumnFallback)
+  end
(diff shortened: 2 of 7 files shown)


lib/mobility/plugins/original_column.rb
```diff
@@ -0,0 +1,15 @@
...
+    module OriginalColumn
+      extend Plugin
+
+      default false
+
+      requires :backend, include: :before
+    end
+
+    register_plugin(:original_column, OriginalColumn)
+  end
```
shioyama, line 14: I haven't really documented this anywhere yet, but this is a kind of convention/trick of defining a global plugin which does nothing, then defining a corresponding ORM-specific plugin which depends on the global one, and which is included automatically when you include the `active_record` plugin. So it's included but disabled by default.

That way, in your plugins, enabling the empty base plugin (`original_column`) indirectly enables the ORM-specific one. If in the future there was a corresponding plugin for Sequel, the same name could be used for that one (this is how `dirty`, `cache` and `query` work).

lib/mobility/plugins/active_record/original_column.rb
```diff
@@ -0,0 +1,60 @@
...
+        included_hook do |_, backend_class|
+          if options[:original_column]
+            backend_class.include BackendInstanceMethods
+            backend_class.extend BackendClassMethods
+          end
+        end
+
+        module BackendInstanceMethods
+          def read(locale, **)
+            if locale == I18n.default_locale
```
shioyama, line 25: I don't love this dependence on `I18n` ... :thinking:

lib/mobility/plugins/active_record/original_column.rb
```diff
@@ -0,0 +1,60 @@
...
+              super
+            end
+          end
+        end
+
+        module BackendClassMethods
+          def build_node(attr, locale)
+            if locale == I18n.default_locale
+              # is the MobilityExpressions plugin necessary here?
+              model_class.arel_table[attr].extend(Plugins::Arel::MobilityExpressions)
```
shioyama, line 45: Never got around to investigating this.

lib/mobility/plugins/original_column.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Mobility
+  module Plugins
+    module OriginalColumn
+      extend Plugin
+
+      default false
```
shioyama, line 8: The default is `false` which means that it needs to be explicitly passed `original_column: true` on every call to `translates`. I think this makes sense since this plugin incurs a perf hit on every read/write, and probaly people only want to use it selectively on some columns/models.

shioyama:
This is actually a really really useful plugin and super simple. Hoping we can get the details worked out and release in 1.2.

lib/mobility/plugins/active_record/original_column.rb
shioyama: I don't _think_ this should be necessary.

lib/mobility/plugins/active_record/original_column.rb
```diff
@@ -0,0 +1,60 @@
...
+            else
+              super
+            end
+          end
+
+          # def apply_scope
+          #   super
+          # end
```
shioyama, line 53: This should be fine to remove; it will only add joins when it sees translated attribute nodes, but since `build_node` will use the column node for the default locale, it should not join for the default locale case.

shioyama:
Note that with the new API for querying with a block, this will work with stuff like:

```ruby
I18n.default_locale = :en
Post.i18n do
  title(:en).eq("foo").or(title(:ja).eq("bar"))
end.to_sql
```

```sql
SELECT "posts".* FROM "posts"
  LEFT OUTER JOIN "mobility_string_translations" "Post_title_ja_string_translations"
    ON "Post_title_ja_string_translations"."key" = 'test'
    AND "Post_title_ja_string_translations"."locale" = 'ja'
    AND "Post_title_ja_string_translations"."translatable_type" = 'Post'
    AND "Post_title_ja_string_translations"."translatable_id" = "posts"."id"
  WHERE ("posts"."title" = 'foo' OR "Post_title_ja_string_translations"."value" = 'bar')
```

Here the backend is KeyValue, but would work for any backend. Notice it's only joining for `ja` and not for `en` since `en` is the default locale.

lib/mobility/plugins/active_record.rb
```diff
@@ -4,6 +4,7 @@
 require_relative "./active_record/cache"
 require_relative "./active_record/query"
 require_relative "./active_record/uniqueness_validation"
+require_relative "./active_record/original_column"
```
shioyama, line 7: We could maybe debate whether this is a good name.

lib/mobility/plugins/active_record/original_column.rb
mattboldt: I actually only used `I18n.default_locale` as an example. In practice, I actually need it to work with `en` OR `en-US`, where `en-US` falls back to plain `en`.

So this check would for sure be configurable -- either a static plugin config, or maybe a proc called in-context when building a new instance.

lib/mobility/plugins/active_record.rb
mattboldt: I was thinking the name could include the word "fallback" or "fallthrough" like with `fallthrough_accessors`, since the plugin only interferes if certain conditions are met. Maybe `fallthrough_original_column`, `original_column_fallthrough`?

lib/mobility/plugins/original_column.rb
mattboldt: Oh! This may be the right spot to get rid of that `I18n` dependency.
```
class Post
  extend Mobility
  translates :name, type: :string, original_column_fallbacks: [:en, :en-US]
end
```

shioyama: Yes indeed! I'll try that.

lib/mobility/plugins/active_record/original_column.rb
shioyama: It could be optionally a proc which would be evaluated in context. The problem is that then we need to use `define_method` instead of `def` which has (small) perf implications, which I try to avoid on anything that runs through attribute getters/setters.

lib/mobility/plugins/active_record.rb
shioyama: Yeah hmm still thinking about it :thinking:

shioyama:
Thanks @stebo for the feedback! This is a great and easy win, so I'd like to give it a name and work out details, and get it out in a 1.2 release soon.

Klockner:
> Thanks @stebo for the feedback! This is a great and easy win, so I'd like to give it a name and work out details, and get it out in a 1.2 release soon.

Hello @shioyama, you still have plans to release this one? Thanks!

shioyama:
Yep! It's planned for the next release.

shioyama:
So ok folks, sorry for the delay on this.

For the name, I'm thinking of just `column_fallback` (singular), so:

```ruby
class Post
  extend Mobility
  translates :name, type: :string, column_fallback: true # falls back on I18n.default_locale
  # translates :name, type: :string, column_fallback: [:en, :fr] # falls back if Mobility.locale is :en of :fr
  # translates :name, type: :string, column_fallback: ->(locale) { ... } # falls back if proc returns true
end
```

WDY(all)T?

peter-tran-goldenowl:
hi @shioyama . I have pull this branch and testing it but can not config it correctly. Could you please provide me a correct config? Here is my current config:

`translates :name,type: : string, original_column: true, original_column_fallbacks: [:en]`

gando001:
Hi @shioyama is there an ETA on the next release including this PR?

shioyama:
Hi folks, haven't had time to get to this. Since it's a new thing (unlike bugs etc) I need to sit down and think about it a bit, so it's slightly downprioritized for me right now.

@peter-tran-goldenowl It's not working since it's not actually implemented yet :sweat_smile: I was just proposing an interface.

shioyama:
Hi folks! Sorry for the delay. I've rewritten this plugin supporting the boolean, locale array and proc options, and added tests. The way it is implemented each option type gets the best performance, so boolean just includes a module, locales array dynamically defines the `read` and `write` methods with `module_eval` (so dispatch is fast) and the proc option defines them with `define_method` (slowest, but unavoidable).

Tests are passing so once y'all try this out and are happy I should be able to release it quickly in a 1.2 release.

mattboldt:
@shioyama that's terrific! Thank you so much for your hard work on this. Gonna try to work this into our team's sprint and get it tested. If all goes well, it'll be in production in ~30 days! 🎉

shioyama:
I added a Sequel implementation, almost identical.

shioyama:
Oh just realized querying is not working for anything other than a `true` option value. I'll fix it and add specs covering that.

shioyama:
Ok added querying for all options in AR::ColumnFallback, with tests to check. Still need to finish other options for Sequel version.

benoror:
heads up/q: is this plugin documented somewhere?

Issue #513, opened 2021-04-09
Query failing with with undefined method `mobility_attribute?`


DJA89:
There's a very very specific query that's failing with that error

`NameError: undefined method 'mobility_attribute?' for class`

the class in question does not extend mobility.

## Context
Here is a repo where you can reproduce this issue https://github.com/DJA89/mobility_issue

## Expected Behavior
The query should work with or without the i18n scope

## Actual Behavior
I have here an example of how the classes are and what specific query is breaking

```ruby
class ParkingLot < ActiveRecord::Base
  belongs_to :car, inverse_of: :parking_lots
  has_many :car_parts, through: :car
end
```
```ruby
class Car < ActiveRecord::Base
  extend Mobility

  translates :some_car_attribute, :some_other_car_attribute

  has_many :parking_lots
  has_many :car_parts

  default_scope { i18n }
  scope :some_car_scope, -> { where(some_thir_car_attribute: true) }
end
```

```ruby
class CarPart < ActiveRecord::Base
  extend Mobility

  translates :some_car_part_attribute

  belongs_to :car, touch: true

  default_scope { i18n }
end
```

Failing query:
```ruby
ParkingLot.includes(car: [:car_parts]).references(:car).merge(Car.some_car_scope).find_by(car_id: 8)
```

If I remove the `[:car_parts]` from the `includes` or the `merge` method from the query then the query works, but exactly as it's shown it fails with `NameError: undefined method 'mobility_attribute?' for class ` and it's trying to call the method for the class `ParkingLot`, the funny thing is the error looks correct, because `ParkingLot` does not extend Mobility, the question is why is it trying to call that method? We aren't even looking for any specific translated field. The query also works if I remove the default scope from the `Car` model.

```
NameError: undefined method `mobility_attribute?' for class `#<Class:0x00007f97a42fcc48>'
/Users/some_user/.rvm/gems/ruby-2.5.5/gems/mobility-1.1.1/lib/mobility/plugins/active_record/query.rb:164:in `method'
/Users/some_user/.rvm/gems/ruby-2.5.5/gems/mobility-1.1.1/lib/mobility/plugins/active_record/query.rb:164:in `block (2 levels) in <module:QueryExtension>'
/Users/some_user/.rvm/gems/ruby-2.5.5/gems/activerecord-5.2.5/lib/active_record/relation/finder_methods.rb:410:in `limited_ids_for'
[...]

PR #515, opened 2021-04-25
Check whether class responds to mobility_attribute?

shioyama:
~This shouldn't be necessary from AR >= 7.0~ _Only one of the two checks is required in AR 7.0_

See: https://github.com/shioyama/mobility/issues/513

### DIFF
--- a/spec/mobility/plugins/active_record/query_spec.rb
+++ b/spec/mobility/plugins/active_record/query_spec.rb
@@ -153,4 +153,43 @@ describe Mobility::Plugins::ActiveRecord::Query, orm: :active_record, type: :plu
       end).to eq([article_ja])
     end
   end
+
+  describe "regression for #513" do
+    plugins :active_record, :query
+    before do
+      m = ActiveRecord::Migration.new
+      m.verbose = false
+
+      m.create_table :cars
+      stub_const('Car', Class.new(ActiveRecord::Base) do
+        has_many :parking_lots
+        has_many :car_parts
+      end)
+      translates Car, backend: :column
+
+      m.create_table(:parking_lots) { |t| t.integer :car_id }
+      stub_const('ParkingLot', Class.new(ActiveRecord::Base) do
+        belongs_to :car
+      end)
+
+      m.create_table(:car_parts) { |t| t.integer :car_id }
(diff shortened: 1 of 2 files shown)


spec/mobility/plugins/active_record/query_spec.rb
```diff
@@ -153,4 +153,44 @@
...
+      m.verbose = false
+      m.drop_table :cars
+      m.drop_table :parking_lots
+      m.drop_table :car_parts
+    end
+
+    it "does not raise NameError" do
+      query = ParkingLot.includes(car: :car_parts).references(:car).merge(Car.some_car_scope)
+      expect { query.first }.not_to raise_error
+      expect { query.order(:car_id) }.not_to raise_error
```
shioyama, line 193: @DJA89 I managed to narrow the problem down a bit, it doesn't even require any actual translated attributes, nor the `through` relation, and only `Car` needs to extend Mobility.

I'm still not really clear what the underlying problem is but since it no longer occurs in Rails 7.0 (master) I'm assuming it's something that's been "fixed", so this will be temporary until Mobility no longer supports versions < Rails 7 (still a long way away!).

shioyama:
The 7.0 failures are a different issue related to dirty method visibility :disappointed:

lib/mobility/plugins/active_record/dirty.rb
```diff
@@ -51,9 +51,11 @@ module Dirty
           end
         end
 
-        included_hook do |_, backend_class|
+        included_hook do |klass, backend_class|
           if options[:dirty]
             backend_class.include BackendMethods
+            # I'm not sure why this is now public, but it is ¯\_(ツ)_/¯
+            klass.send(:public, :changes_applied) if ::ActiveRecord::VERSION::STRING >= '7.0'
```
shioyama, line 58: This has nothing to do with this PR, but specs on AR 7.0 are failing unless this is fixed, so sneaking it in.

shioyama: Ok here's what changed this: https://github.com/rails/rails/pull/41860

shioyama: Ok this is a Rails problem, I'll wait for it to be fixed and rebase. We shouldn't need this patch.

Issue #514, opened 2021-04-21
Removing presence does not work for jsonb backend


DJA89:
Commenting out the `presence` option on the initializer does not work, at least for the `jsonb` backend

## Context

I commented out `presence` from the initializer, but when trying to save or update a model with an empty string, it still saves them as `nil`.

Using the same repo as the other issue but on a different branch as an example.
https://github.com/DJA89/mobility_issue/tree/without_presence

## Expected Behavior
I should be able to save empty strings on translated attributes

## Actual Behavior
It transforms empty strings to `nil` values

## Possible Fix
I don't know of a possible fix.

shioyama:
If you comment out a plugin, it absolutely removes it (doesn't even load it, the constant `Mobility::Plugins::Presence` will not even be defined).

The case here is specific to the jsonb plugin, I think this line:

https://github.com/shioyama/mobility/blob/ef25f9a55fa3c26f3bbcd09170f6ec14b3047907/lib/mobility/backends/active_record/pg_hash.rb#L35

Maybe that `if value.present?` isn't needed, but I have a feeling there were other reasons that it is there. Why don't you try patching that line and run against the patched version, see how it works for you?

shioyama:
Also, that change would impact `hstore` and `json` in addition to `jsonb` (the same code is used by all three).

SleeplessByte:
I've ran into this where `{}` (no translations) is mapped to `nil` instead (postgresql, jsonb column backend). It seems to be caused by `pg_hash.rb` indeed.

DJA89:
Hey! We so far have decided to try and work around this, accepting the `nil` values or trying to use the default values when that's not possible. 

We chose the "error" that we know is there instead of finding out what were the other reasons you had decided to add that condition and break everything 😅 .

Thanks!

shioyama:
Ok #516 is passing (updated) tests now. I changed the code to only set the value if it is non-nil, rather than present (i.e. allow blank values), and all tests other than the ones that were explicitly checking blank value serialization pass. So I think this is a good change :+1: 

I'll probably release it as a minor bump to 1.2

DJA89:
Thank you! That's awesome! :D

shioyama:
Ok this is merged to `master`, please point your `Gemfile` to that for now. I'll release 1.2 soon-ish probably with some other changes in the pipeline.

PR #516, opened 2021-05-07
Set value even if blank in pg_hash

shioyama:
Related: #514


shioyama:
Tests will fail, but I'd like to see which ones and how they fail.

shioyama:
I just need to clean up these commits, but otherwise this seems like a reasonable change. I _think_ when I originally wrote this code the "presence" plugin wasn't a plugin yet (because there weren't things such as plugins yet back then), but now that there are the presence check in the serializer(s) is redundant and unnecessary.

PR #520, opened 2021-07-23
Do not swallow keyword args on ruby 3 in fallthrough accessors

doits:
If a `method_missing` is defined explicitly with kwargs in own code, for example like this:

```ruby
        def method_missing(method_name, *args, **kwargs, &block)
           ...
        end
```

*and* fallthrough accessors are used, it looks like `kwargs` are swallowed by mobility's definition of `method_missing`.

For now I only added a failing spec to illustrated what I think should happen.

The spec **passes on ruby 2.7** but **fails on ruby 3**.

I'm not sure what the solution is for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
+          - '3.0'
           - '2.7'
           - '2.6'
         database:
@@ -45,6 +46,10 @@ jobs:
         experimental: [false]
         feature: ['unit']
         include:
+          - ruby: '3.0'
+            feature: 'unit'
+            orm:
+            experimental: false
           - ruby: '2.7'
             feature: 'unit'
             orm:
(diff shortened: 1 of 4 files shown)


doits:
I updated CI to run specs with ruby 3 so you can see the failing spec there.

doits:
I think I fixed it by using `ruby2_keywords`, which is the way suggested [by the ruby team](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/#a-compatible-delegation).

If support for ruby <= 2.6 is dropped they suggest to use `**kwargs` explicitly, but until then `ruby2_keywords` and *not* using `**kwargs` seems to be the only solution to be compatible with all versions.

Actually, since `**kwargs` is used in mobility [like here](https://github.com/shioyama/mobility/blob/master/lib/mobility/plugins/fallthrough_accessors.rb#L51), doesn't it implicitly mean that mobility is only compatible with ruby >= 2.7? (I think this syntax is only valid for ruby >= 2.7). If yes, then the `method_missing` might be rewritten without `ruby2_keywords`.

shioyama:
Thanks @doits ! Really appreciate this. Can you squash commits?

shioyama (approved):

doits:
:+1: squashed it, I think this is ready now.

shioyama:
Thanks! Released in 1.1.3.

PR #525, opened 2021-08-30
Configurable KeyValue backend's translation classes

thedarkside:
The [mobility-actiontext](https://github.com/sedubois/mobility-actiontext) implementation is broken because the KeyValue backend (which is being used) is coupled to the internal translation classes. See here for the issue: https://github.com/sedubois/mobility-actiontext/issues/3

With this pullrequest the coupling is loosened up by making it configurable. The previous behavior is still the default to prevent breaking things.

Once this pullrequest is merged, https://github.com/sedubois/mobility-actiontext/pull/12 can be merged too!

### DIFF
--- a/lib/mobility/backends/active_record/key_value.rb
+++ b/lib/mobility/backends/active_record/key_value.rb
@@ -34,6 +34,7 @@ Implements the {Mobility::Backends::KeyValue} backend for ActiveRecord models.
         def configure(options)
           super
           if type = options[:type]
+            options[:include_other_translation_classes] ||= Mobility::Backends::ActiveRecord::KeyValue::Translation.descendants.uniq
             options[:association_name] ||= :"#{options[:type]}_translations"
             options[:class_name]       ||= const_get("#{type.capitalize}Translation")
           end
@@ -155,6 +156,7 @@ Implements the {Mobility::Backends::KeyValue} backend for ActiveRecord models.
         key_column         = options[:key_column]
         value_column       = options[:value_column]
         belongs_to         = options[:belongs_to]
+        include_other_translation_classes = options[:include_other_translation_classes]
 
         # Track all attributes for this association, so that we can limit the scope
         # of keys for the association to only these attributes. We need to track the
@@ -192,7 +194,7 @@ Implements the {Mobility::Backends::KeyValue} backend for ActiveRecord models.
         end
 
         # Ensure we only call after destroy hook once per translations class
-        translation_classes = [translation_class, *Mobility::Backends::ActiveRecord::KeyValue::Translation.descendants].uniq
+        translation_classes = [translation_class, *include_other_translation_classes].uniq
         after_destroy do
           @mobility_after_destroy_translation_classes = [] unless defined?(@mobility_after_destroy_translation_classes)
           (translation_classes - @mobility_after_destroy_translation_classes).each { |klass| klass.where(belongs_to => self).destroy_all }


lib/mobility/backends/active_record/key_value.rb
```diff
@@ -34,6 +34,7 @@ class << self
         def configure(options)
           super
           if type = options[:type]
+            options[:include_other_translation_classes] ||= Mobility::Backends::ActiveRecord::KeyValue::Translation.descendants.uniq
```
sedubois, line 37: How about a boolean flag instead?

```rb
# Also, don't read the option within `if type = options[:type]`, 
# as it appears that `Mobility::Backends::ActionText` does not pass an `options[:type]`. 
# https://github.com/sedubois/mobility-actiontext/blob/382016bc43148e079b2652021e5d5bf982c56c75/lib/mobility/backends/action_text.rb#L42-L53

options[:destroy_descendants] = options.fetch(:destroy_descendants, true)
```

lib/mobility/backends/active_record/key_value.rb
```diff
@@ -155,6 +156,7 @@ def visit_default(_)
         key_column         = options[:key_column]
         value_column       = options[:value_column]
         belongs_to         = options[:belongs_to]
+        include_other_translation_classes = options[:include_other_translation_classes]
```
sedubois, line 159:
```suggestion
        destroy_descendants = options[:destroy_descendants]
```

lib/mobility/backends/active_record/key_value.rb
```diff
@@ -192,7 +194,7 @@ def visit_default(_)
         end
 
         # Ensure we only call after destroy hook once per translations class
-        translation_classes = [translation_class, *Mobility::Backends::ActiveRecord::KeyValue::Translation.descendants].uniq
+        translation_classes = [translation_class, *include_other_translation_classes].uniq
```
sedubois, line 197:
```suggestion
        translation_classes = if destroy_descendants
          [translation_class, *Mobility::Backends::ActiveRecord::KeyValue::Translation.descendants].uniq
        else
          [translation_class]
        end
```

sedubois (changes requested):

lib/mobility/backends/active_record/key_value.rb
sedubois: @shioyama is probably the best person to answer those questions 🙂

shioyama: Sorry for the late reply!

> Maybe we should first clarify the reason why its necessary to take the other Translation classes other then the one inside translation_class into account while destroying?

Yes I admit this is quite hard to understand from the code.

The reason why the code attempts to destroy _all_ translation type records is that you can start by declaring (say) a `text` translation type and then later switch to `string`, and you will leave `text` translation orphans attached to the model record.

So the code is intended to guarantee that when the model is destroyed, all translations attached to it, regardless of their type, will also get destroyed.

TBH, looking at this code now (a couple years after I wrote it), I'm not sure that the complexity tradeoff is worth it. But OTOH it _is_ very easy to change translation types and leave behind orphaned translations.

Anyway, I don't really want to change the default, but I agree it should be possible to disable it. My only question is: should the option be to _completely_ disable the `after_destroy` hook, or disable destroying the descendants (as you have done here) ?

I'm a bit concerned that someone may come along later and want to implement the whole `after_destroy` thing differently, at which point we'd need yet another option for this. If that's the case, maybe we should just have one option which disables the whole thing (so no cleanup of translations of any kind) and anyone who uses it has to manage translations cleanup on their own.

WDYT?

shioyama:
I agree with the intent of this PR to make the `after_destroy` logic more customizable, my main concern is that if we are going to add yet another option to the `KeyValue` backend, it should be enough that we don't need another one (for this destroy behaviour at least).

So I'd suggest an option named something like `destroy_translations`, which defaults to `true`, but which if `false` completely disables the _entire_ `after_destroy` hook. I think this is more transparent and means that people can completely customize what the destroy code does. Then in the ActionText gem you can just implement whatever logic is best, and not have to worry about how Mobility defines it.

sedubois:
I'm OK with Chris' proposal. @thedarkside?

shioyama:
@thedarkside 

> While this may be a good feature it does not solve my issue here.

I agree, you're right that for the case in mobility-actiontext this is not an ideal solution. What you really want is to just override the entire action of destroying translations, not change it through an option value.

Another way to do this is to define a class method on the backend which is called from inside the `setup` block. The method calls `after_destroy`, so basically nothing changes, _except_ that now that it is a backend class method it can be overridden in subclasses.

See ceb3a1f8fcec9b3104f80f5f18224d1f7a86bfd7 (feel free to copy/use this here).

I think this is perhaps the best solution for this problem. Then mobility-actiontext can simply override this method and do whatever is best for the gem.

The only thing then is to be sure that this method is well-defined since it needs to be public and so should not change in the future. We can discuss whether what I proposed in ceb3a1f8fcec9b3104f80f5f18224d1f7a86bfd7 is a good choice or not.

shioyama:
A bit tangential, but this discussion got me thinking that it's important to draw a line between "base backend classes" (including subclasses of them like the one in mobility-actiontext discussed here) and the very specific subclasses that are used inside of models. I'm making that distinction clearer in https://github.com/shioyama/mobility/pull/527

shioyama:
Ok I think #528 should do the trick. Now `setup` blocks take an optional third argument, the configured backend class. That makes it easy to define class methods on the backend class which do the work of steps inside the setup block, which also allows subclasses to override them.

So if #528 works, then you just need to override `define_after_destroy_callback` on the mobility-actiontext subclass to do whatever you want. No option needed, and even if we added one in the future since you'd be overriding the method completely it wouldn't impact the subclass.

thedarkside:
@shioyama thank you for your effort. Makes sense and works great!

We can close this in favor of https://github.com/shioyama/mobility/pull/528 

Im eagerly looking forward to those changes being released ;)

PR #526, opened 2021-09-21
Registerable translation classes for subclassing plugins

thedarkside:
Its an alternative proposal to https://github.com/shioyama/mobility/pull/525 allowing to 
- utilize the destroy mechanism already implemented by the key_value backend
- allow subclassing plugins to flexibly change the registered translation classes (adding, deleting)
- not polluting options

Also it should work nicely when mixing backends for different attributes in the same model (key_value and action_text together)

### DIFF
--- a/lib/mobility/backends/active_record/key_value.rb
+++ b/lib/mobility/backends/active_record/key_value.rb
@@ -165,6 +165,10 @@ Implements the {Mobility::Backends::KeyValue} backend for ActiveRecord models.
         association_attributes = (instance_variable_get(:"@#{attrs_method_name}") || []) + attributes
         instance_variable_set(:"@#{attrs_method_name}", association_attributes)
 
+        translation_classes_method_name = :"__#{association_name}_translation_classes"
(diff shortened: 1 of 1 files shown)


lib/mobility/backends/active_record/key_value.rb
```diff
@@ -165,6 +165,10 @@ def visit_default(_)
         association_attributes = (instance_variable_get(:"@#{attrs_method_name}") || []) + attributes
         instance_variable_set(:"@#{attrs_method_name}", association_attributes)
 
+        translation_classes_method_name = :"__#{association_name}_translation_classes"
+        association_translation_classes = (instance_variable_get(:"@#{translation_classes_method_name}") || []) + Mobility::Backends::ActiveRecord::KeyValue::Translation.descendants
+        instance_variable_set(:"@#{translation_classes_method_name}", association_translation_classes)
```
shioyama, line 170: I'd really prefer to avoid setting new instance variables in the model class. I know it's done in other places but it's not a very good pattern, and I think there are better ways we can achieve the same thing. I'll comment on #525.

thedarkside:
You are right. Closing this in favor of https://github.com/shioyama/mobility/pull/528

PR #527, opened 2021-09-23
Clearly distinguish backend classes from their configured subclasses

shioyama:
In practice, every backend class that is used in a model is a subclass of one of the base classes like `Mobility::Backends::ActiveRecord::KeyValue` etc. Configured backend classes have a model class (points to the class in which the backend is used) and set of options configuring it, whereas the base backend classes do not.

However, previously you could still call `options` or `model_class` on the base classes and you would get back a `nil`. This is confusing since those getter methods should never be called on the base backend classes.

In this commit I'm restructuring the code such that:
- getter methods raise on the base backend class
- methods which do not make sense on the base backend class are moved to a module which is included into backend subclasses.

I think this should avoid weird bugs and misunderstandings.

### DIFF
--- a/lib/mobility/backend.rb
+++ b/lib/mobility/backend.rb
@@ -117,7 +117,6 @@ On top of this, a backend will normally:
     # Extend included class with +setup+ method and other class methods
     def self.included(base)
       base.extend ClassMethods
-      base.singleton_class.attr_reader :options, :model_class
     end
 
     # Defines setup hooks for backend to customize model class.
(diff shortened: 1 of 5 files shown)


lib/mobility/backend.rb
```diff
@@ -147,17 +146,6 @@ def setup &block
 
       def inherited(subclass)
         subclass.instance_variable_set(:@setup_block, @setup_block)
-        subclass.instance_variable_set(:@options, @options)
-        subclass.instance_variable_set(:@model_class, @model_class)
```
shioyama, line 151: `@options` and `@model_class` were being assigned to subclasses, but this doesn't really make sense since:
- "base" backend classes don't have these, so their subclasses are just assigning nil values
- configured backend classes don't (or shouldn't) have subclasses, so this would not be called anyway

PR #528, opened 2021-09-23
Allow passing configured backend class as third argument to setup

shioyama:
Refactoring the `setup` block in backend classes can be difficult because the local context in the block is the model class, and we don't want to dump methods there.

This PR makes the `setup` block accept an optional third argument, the configured backend class, which can then be the receiver for method calls to extract common logic (or customizable logic) out of the `setup` block. An immediate use case for this is #525.

### DIFF
--- a/lib/mobility/backend.rb
+++ b/lib/mobility/backend.rb
@@ -21,8 +21,8 @@ On top of this, a backend will normally:
   corresponding to valid keys for configuring this backend.
 - implement a +configure+ class method to apply any normalization to the
   keys on the options hash included in +valid_keys+
-- call the +setup+ method yielding attributes and options to configure the
-  model class
+- call the +setup+ method yielding attributes and options (and optionally the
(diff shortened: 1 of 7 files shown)


thedarkside:
Okay cool! Ive adjusted the https://github.com/sedubois/mobility-actiontext/pull/12 to make use of those changes.

shioyama:
Ok I went ahead and extracted everything else out of the `setup` block for KeyValue (AR + Sequel) as well. I won't do it for other backends yet but the pattern is now there so it could be done.

Once tests pass I'll merge this.

shioyama:
@thedarkside I changed the method signature: it takes the model class now, and nothing else. I'll merge this, please work against what's in the master branch.

shioyama:
I just released 1.2 which includes this PR.

thedarkside:
@shioyama cool, thank you. Will do!

PR #532, opened 2021-10-02
Make models work with `Marshal.dump` (take two)

shioyama:
Replaces #505

~I figured out how to avoid an md5 digest or other hack when naming configured backends. Basically, we don't name them based on their attributes but instead on the attribute names in the model that the backend class is created for. This is cleaner and means the backend class name is directly readable, so less unintuitive.~

~I've also assigned the name to a namespace under the backend class itself, rather than off of the root namespace. This way the name only has to incorporate the model name and attribute name(s).~

_Update: I simplified this by just avoiding dumping the backends at all (by defining `marshal_dump` and `marshal_load`), which gets around the problem entirely._

### DIFF
--- a/lib/mobility/backend.rb
+++ b/lib/mobility/backend.rb
@@ -77,6 +77,12 @@ On top of this, a backend will normally:
       @attribute = args[1]
     end
 
+    def ==(backend)
+      backend.class == self.class &&
+        backend.attribute == attribute &&
+        backend.model == model
+    end
+
     # @!macro [new] backend_reader
     #   Gets the translated value for provided locale from configured backend.
     #   @param [Symbol] locale Locale to read
--- a/lib/mobility/plugins/backend.rb
+++ b/lib/mobility/plugins/backend.rb
@@ -114,15 +114,25 @@ Defines:
         defaults[key] = [backend, backend_options] if backend
       end
 
+      class MobilityBackends < Hash
+        def initialize(model)
+          @model = model
+          super()
+        end
+
+        def [](name)
+          return fetch(name) if has_key?(name)
+          return self[name.to_sym] if String === name
+          self[name] = @model.class.mobility_backend_class(name).new(@model, name.to_s)
+        end
+      end
+
       module InstanceMethods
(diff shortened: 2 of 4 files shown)


shioyama:
So for a model class `Article::Foo`, and an attribute `title` with the null backend, this becomes:

```ruby
backend_class_1
#=> #<Mobility::Backends::Null>
backend_class_1.name
#=> "Mobility::Backends::Null::Article_Foo__title"
```

So `inspect` shows something intuitive (`#<Mobility::Backends::Null>`), but `name` (which is the actual constant) is also fairly intuitive (`"Mobility::Backends::Null::Article_Foo__title"`).

For backends defined on multiple attribute together, the name becomes e.g. `"Mobility::Backends::Hash::Article_Foo__foo_bar"`.

If attribute names change, `Marshal.load` will fail. However, if _options_ change, it will still work. This is different from #505 where changes to options would break load but attribute name changes would not.

I'm not sure which is better but options are much harder to cleanly convert to a name, so I think this is the better approach. In any case `Marshal.load` is risky so anyone using it should always prepare for failures to deserialize after code changes (cache reset etc).

shioyama:
Looks like Sequel postgres tests are failing (not related to this). I'll look at that in a separate PR.

shioyama:
@maxime-carbonneau I think this is a better solution. I just need to fix that Sequel thing and then this should be good to go.

shioyama:
Ok I found an even easier solution: don't dump the backends, and just regenerate them after reloading. This avoids all the problems. So so much easier.

maxime-carbonneau:
@shioyama Thank you

cspeer:
Hi @shioyama,

should this change make using Mobility with Rails.cache work?
Unfortunately, I'm still getting a `can't dump anonymous class` error when I try to cache localized objects:
```ruby
@articles = Rails.cache.fetch({ locale: I18n.locale, key: 'articles' }, expires_in: 6.hours) do
  Article.i18n.order(name: :asc)
end
```
`name` is an attribute controlled by Mobility.

Am I missing something?

PR #536, opened 2021-10-05
Fix ActiveRecord JSONB blank values

### LINKED ISSUE #535: JSONB + empty default bug
When a JSONB column has empty default (`{}`) [as suggested](https://github.com/shioyama/mobility/wiki/Postgres-Backends-(Column-Attribute)#railsactiverecord), and it is wrapped with Mobility `translates`, reading from that column messes up ActiveRecord's dirty tracking. After you read from that column, `changes` start reporting very confusing diff (no writes were performed):

```ruby
a_record.changes # => {"a_column_i18n" => [{}, {}]}
```

This is not related to the [Mobility's dirty tracking plugin](https://github.com/shioyama/mobility#dirty-tracking) as removing/disabling it does not fix the behavior.

## Context
https://gist.github.com/nashbridges/764f5da35d1a4f13eb98fccf647ef8b2

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails", "6.1.4.1"
  gem "pg", "1.2.3"
  gem "mobility", "1.2.2"
end

require "active_record"
require "minitest/autorun"
require "logger"
require "mobility"

Mobility.configure do
  plugins do
    active_record
  end
end

ActiveRecord::Base.establish_connection(adapter: "postgresql", database: "mobility_test")
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :properties, force: true do |t|
    t.jsonb :name_i18n, default: {}, null: false
    t.jsonb :long_name_i18n
  end
end

class PropertyWithoutMobility < ActiveRecord::Base
  self.table_name = "properties"
end

class PropertyWithMobility < ActiveRecord::Base
  self.table_name = "properties"
  extend Mobility
  translates :name, backend: :jsonb, column_suffix: "_i18n"
  translates :long_name, backend: :jsonb, column_suffix: "_i18n"
end

class BugTest < Minitest::Test
  def test_without_mobility_reading_from_not_null_column
    property = PropertyWithoutMobility.create!
    property.reload
    assert_equal({}, property.name_i18n)
    assert_equal({}, property.changes)
  end

  def test_without_mobility_reading_from_null_column
    property = PropertyWithoutMobility.create!
    property.reload
    assert_nil property.long_name_i18n
    assert_equal({}, property.changes)
  end

  def test_with_mobility_reading_from_not_null_column
    property = PropertyWithMobility.create!
    property.reload
    assert_equal({}, property.name_i18n)
    assert_equal({}, property.changes)
  end

  def test_with_mobility_reading_from_null_column
    property = PropertyWithMobility.create!
    property.reload
    assert_nil property.long_name_i18n
    assert_equal({}, property.changes)
  end
end
```

We use Postgres 13.2.

## Expected Behavior
All tests should pass.

## Actual Behavior
I have two failing tests:

```
Failure:
BugTest#test_with_mobility_reading_from_not_null_column [jsonb_blank.rb:63]:
Expected: {}
  Actual: {"name_i18n"=>[{}, {}]}
```

```
Failure:
BugTest#test_with_mobility_reading_from_null_column [jsonb_blank.rb:69]:
Expected {} to be nil
```

The second failure is interesting, because it shows that by _not_ providing the `{}` default for the DB column we can get the desired behavior: there's no `nil` related errors + `changes` is not mutated. For now this is what we do in our application (dropping the default) as a workaround.

## Possible Fix

I was able to track down the source of the issue to different column type.

When there's no `translates` the JSONB column is handled by the [JSONB type](https://github.com/rails/rails/blob/c2b701e33470adb1fab15c5e68957facdb26ebb1/activerecord/lib/active_record/connection_adapters/postgresql/oid/jsonb.rb), which treats the default correctly.
When the column is wrapped with `translates` it is handled by the [Serialized type](https://github.com/rails/rails/blob/c2b701e33470adb1fab15c5e68957facdb26ebb1/activerecord/lib/active_record/type/serialized.rb), which somehow [returns `nil` for a default value case](https://github.com/rails/rails/blob/c2b701e33470adb1fab15c5e68957facdb26ebb1/activerecord/lib/active_record/type/serialized.rb#L64).

It's not clear though what would be the fix.

shioyama:
Fixes #535

This passes the regression test, but maybe we can simplify it.


shioyama:
Although the tests never checked this, there is a breaking change here in that the translations column (`title_i18n` etc) goes from being an `ActiveSupport::HashWithIndifferentAccess` to being just a plain `Hash` with string keys.

For most people, this should make no difference, but I can imagine cases where people are doing some kind of custom stuff passing the locale to the hash to get translations, which will now break unless the locale is converted to a string first.

I'll release this as part of 1.3 rather than a patch for that reason, with a warning in the changelog entry.

PR #543, opened 2021-11-05
clean up and refactor active record container backend

### LINKED ISSUE #540: Backend container reader mixes up dirty tracking and creates empty locale hash
This is in a similar vein to #535: The container backend does some weird things to the dirty tracking after a reload or after accessing a reader.

One is about `.changes` where it states that the column changed like this: `{"translations"=>[{}, {}]}` simply after reload.

```rb
property.translations
# => {}

property.changes
# => {}

property.reload

property.changes
# => {"translations"=>[{}, {}]}
# Expected: {}
```

Second one is that when a reader is accessed it adds the locale key to the translations, but with empty hash:

```rb
property.translations
# => {}

property.changes
# => {}

property.name
# => nil

property.translations
# => {"en"=>{}}
# Expected: {}

property.changes
# => {"translations"=>[{}, {"en"=>{}}]}
# Expected: {}
```

## Bug template

```rb
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails", "6.1.4.1"
  gem "pg", "1.2.3"
  gem "mobility", github: 'shioyama/mobility'
end

require "active_record"
require "minitest/autorun"
require "logger"
require "mobility"

Mobility.configure do
  plugins do
    backend :container

    reader

    active_record
  end
end

ActiveRecord::Base.establish_connection(adapter: "postgresql", database: "mobility_test")
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :properties, force: true do |t|
    t.jsonb :translations, default: {}, null: false
  end
end

class PropertyWithMobility < ActiveRecord::Base
  self.table_name = "properties"
  extend Mobility
  translates :name
end

class BugTest < Minitest::Test
  def test_translations_and_changes_are_blank_after_create_reload
    property = PropertyWithMobility.create!
    property.reload

    assert_equal({}, property.translations, "translations should be blank")
    assert_equal({}, property.changes, "changes should be blank")
    assert_equal(false, property.changed?, "changed should be false")
  end

  def test_translations_and_changes_are_blank_after_accessor
    property = PropertyWithMobility.create!

    property.name

    assert_equal({}, property.translations, "translations should be blank")
    assert_equal({}, property.changes, "changes should be blank")
    assert_equal(false, property.changed?, "changed should be false")
  end
end
```

doits:
fixes dirty tracking and creating empty translation hashes on read

fixes #540

### DIFF
--- a/lib/mobility/backends/active_record/container.rb
+++ b/lib/mobility/backends/active_record/container.rb
@@ -81,8 +81,8 @@ Implements the {Mobility::Backends::Container} backend for ActiveRecord models.
 
       # @!macro backend_iterator
       def each_locale
-        model[column_name].each do |l, v|
-          yield l.to_sym if v.present?
+        model[column_name].each_key do |l|
+          yield l.to_sym
         end
       end
 
@@ -108,22 +108,23 @@ Implements the {Mobility::Backends::Container} backend for ActiveRecord models.
       private
 
       def model_translations(locale)
-        translations = model[column_name]
-
-        return unless translations
-
-        translations[locale.to_s]
+        model[column_name][locale.to_s]
       end
 
       def set_attribute_translation(locale, value)
         locale_translations = model_translations(locale)
 
         if locale_translations
-          locale_translations[attribute.to_s] = value
-          locale_translations.compact!
-          model[column_name].compact!
-        elsif value
-          (model[column_name] ||= {})[locale.to_s] = { attribute.to_s => value }
+          if value.nil?
+            locale_translations.delete(attribute.to_s)
+
+            # delete empty locale hash if last attribute was just deleted
+            model[column_name].delete(locale.to_s) if locale_translations.empty?
+          else
+            locale_translations[attribute.to_s] = value
+          end
+        elsif !value.nil?
+          model[column_name][locale.to_s] = { attribute.to_s => value }
         end
       end
 
--- a/lib/mobility/backends/active_record/container.rb
+++ b/lib/mobility/backends/active_record/container.rb
@@ -22,7 +22,11 @@ Implements the {Mobility::Backends::Container} backend for ActiveRecord models.
       # @param [Hash] options
       # @return [String,Integer,Boolean] Value of translation
       def read(locale, _ = nil)
-        model_translations(locale)[attribute]
+        locale_translations = model_translations(locale)
+
+        return unless locale_translations
+
+        locale_translations[attribute.to_s]
       end
 
       # @note Translation may be a string, integer, boolean, hash or array
@@ -33,7 +37,7 @@ Implements the {Mobility::Backends::Container} backend for ActiveRecord models.
       # @return [String,Integer,Boolean] Updated value
       def write(locale, value, _ = nil)
         set_attribute_translation(locale, value)
-        model_translations(locale)[attribute]
+        read(locale)
       end
       # @!endgroup
 
@@ -42,8 +46,7 @@ Implements the {Mobility::Backends::Container} backend for ActiveRecord models.
         # @option options [Symbol] column_name (:translations) Name of column on which to store translations
         # @raise [InvalidColumnType] if the type of the container column is not json or jsonb
         def configure(options)
-          options[:column_name] ||= :translations
-          options[:column_name] = options[:column_name].to_sym
+          options[:column_name] = options[:column_name]&.to_sym || :translations
         end
         # @!endgroup
 
@@ -84,8 +87,6 @@ Implements the {Mobility::Backends::Container} backend for ActiveRecord models.
       end
 
       setup do |_attributes, options|
-        store options[:column_name], coder: Coder
-
         # Fix for duping depth-2 jsonb column in AR < 5.0
         if ::ActiveRecord::VERSION::STRING < '5.0'
           column_name = options[:column_name]
@@ -107,32 +108,22 @@ Implements the {Mobility::Backends::Container} backend for ActiveRecord models.
       private
 
       def model_translations(locale)
-        model[column_name][locale] ||= {}
+        translations = model[column_name]
+
+        return unless translations
+
+        translations[locale.to_s]
       end
 
       def set_attribute_translation(locale, value)
-        translations = model[column_name] || {}
-        translations[locale.to_s] ||= {}
-        translations[locale.to_s][attribute] = value
-        model[column_name] = translations
-      end
+        locale_translations = model_translations(locale)
 
-      class Coder
-        def self.dump(obj)
-          if obj.is_a? ::Hash
-            obj.inject({}) do |translations, (locale, value)|
-              value.each do |k, v|
-                (translations[locale] ||= {})[k] = v if v.present?
-              end
-              translations
-            end
-          else
-            raise ArgumentError, "Attribute is supposed to be a Hash, but was a #{obj.class}. -- #{obj.inspect}"
-          end
-        end
-
-        def self.load(obj)
-          obj
+        if locale_translations
+          locale_translations[attribute.to_s] = value
+          locale_translations.compact!
+          model[column_name].compact!
+        elsif value
+          (model[column_name] ||= {})[locale.to_s] = { attribute.to_s => value }
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/mobility/backends/active_record/container.rb
```diff
@@ -107,32 +108,22 @@ def initialize_dup(source)
...
 
-        def self.load(obj)
-          obj
+      def set_attribute_translation(locale, value)
+        locale_translations = model_translations(locale)
+
+        if locale_translations
+          locale_translations[attribute.to_s] = value
+          locale_translations.compact!
+          model[column_name].compact!
```
doits, line 124: not sure if those two compacts impact performance or could be optimized

spec/mobility/backends/active_record/container_spec.rb
```diff
@@ -20,6 +20,19 @@
...
+
+      expect { post.title }.not_to change { post.translations }.from({})
+    end
+
+    it 'does not mix up dirty tracking on access' do
+      post = ContainerPost.new
+
+      expect { post.title }.not_to change { post.changes }.from({})
+      expect(post.changed?).to be(false)
+    end
```
doits, line 35: is there a better place for this specs? should they be some shared example for every backend?

shioyama:
Thanks for this! :heart: Will review this week.

lib/mobility/backends/active_record/container.rb
```diff
@@ -107,32 +108,22 @@ def initialize_dup(source)
...
-              value.each do |k, v|
-                (translations[locale] ||= {})[k] = v if v.present?
-              end
-              translations
-            end
-          else
-            raise ArgumentError, "Attribute is supposed to be a Hash, but was a #{obj.class}. -- #{obj.inspect}"
-          end
-        end
+        translations[locale.to_s]
```
shioyama, line 115: We don't need to be defensive here: the default `{}` should be on the column, so this can simply be:

```ruby
model[column_name][locale.to_s]
```

lib/mobility/backends/active_record/container.rb
shioyama: I think it's enough to delete the key if `value` is not `present?`, and otherwise set it. `compact!` assumes we want to clean up other blank translations but that shouldn't be necessary.

so:

```ruby
if locale_translations
  if value.nil?
    locale_translations.delete(attribute.to_s)
  else
    locale_translations[attribute.to_s] = value
  end
end
```

None of these edge cases, mind you, is tested, much or at all. It might make sense to add a container spec which checks that when you set a value of `nil` to an attribute with the container backend, and reload, it comes back `nil`, _and_ the stored `translations` hash has no key for that locale/attribute.

The handling of `""` in the current implementation is wrong. `""` should be treated as a value; if someone wants blank strings treated as `nil` then they can use the presence plugin to do that.

spec/mobility/backends/active_record/container_spec.rb
shioyama: I think this is ok. This is just for general dirty tracking, correct? Not the Mobility dirty plugin. In which case it's a test that applies to any set of plugins using the container backend.

lib/mobility/backends/active_record/container.rb
doits: Doesn't `.compact` only clear `nil` values, so blank strings would not be affected by it and saved correctly?

```rb
{ 'a' => nil, 'b' => "" }.compact!
# => {"b"=>""}
```

Though if we assume the translation column is well formed (has no dangling empty hashes), it could indeed be done like you suggested without `compact` but with one addition to remove empty locale hashes (so we don't save `{ "en" => {} }` but `{}` after deleting the last translation):

```rb
if locale_translations
  if value.nil?
    locale_translations.delete(attribute.to_s)

    # delete empty locale hashes if we removed the last translation
    model[column_name].delete(locale.to_s) if locale_translations.empty?
  else
    locale_translations[attribute.to_s] = value
  end
else 
  ...
end
```

I'll update the implementation and add some specs like you requested (so you can play around with it more easily, too).

doits:
I rebased everything and updated as requested:

- Always assume `model[column_name]` is a hash
- No `compact` anymore
- One spec about removing empty locale hashes when last attribute is deleted

One thing to keep in mind is that empty hashes might already exist in saved models if the container backend was used before this PR and it will not "self heal" (like it did with `compact`). Not sure what to do about it though, or if there should be done anything about it.

doits:
Hey @shioyama, hope you're good! Any chance to review this PR again? I'd like to get this merged and off my list :-) If there are some changes needed I'll update the PR, just say so.

shioyama:
@doits Thanks so much, this is a really great PR! :heart: 

> One thing to keep in mind is that empty hashes might already exist in saved models if the container backend was used before this PR and it will not "self heal" (like it did with compact). Not sure what to do about it though, or if there should be done anything about it.

I think that's ok, if it hasn't been a problem so far then presumably it won't suddenly start being one (for those using the container backend).

lib/mobility/backends/active_record/container.rb
```diff
@@ -33,7 +37,7 @@ def read(locale, _ = nil)
       # @return [String,Integer,Boolean] Updated value
       def write(locale, value, _ = nil)
         set_attribute_translation(locale, value)
-        model_translations(locale)[attribute]
+        read(locale)
```
dramalho, line 40: hi guys :) 

I think there's an issue here - the TLDR is that `set_attribute_translation` isn't touching the cache and *if* the cache has previously been inicialized (say with `{ en: nil }`) , the `read(locale)` will always return `nil` and not the value that we *JUST* wrote 


````ruby
Loading development environment (Rails 8.0.1)
lovely-app(dev)> a = Badge.new
=>
#<Badge:0x00000001158dc4b0
...
lovely-app(dev)> a.title
=> nil
lovely-app(dev)> a.title = "aaa"
=> "aaa"
lovely-app(dev)> a.title
=> nil
```

of course if I *don't* call the reader before, it's fine

```ruby
Loading development environment (Rails 8.0.1)
lovely-app(dev)> a = Badge.new
=>
#<Badge:0x00000001168e41f0
...
lovely-app(dev)> a.title = "aaaaa"
=> "aaaaa"
lovely-app(dev)> a.title
=> "aaaaa"
```

dramalho: the problem is - I think - the `container#write` calls `read` for it's return value, which ends up in `cache#read` -- that ends up returning the previous value (in this case `nil`) and .. deep breath .. the `cache#write` isn't given a chance to rewrite the cached value as it's always getting the same value it has :)

dramalho: What I don't understand is why call `read` when we have the value right there -- I'll look at the commits , maybe there's a hint

dramalho: a little background too, we've been using `mobility` for a while (years) and this particular model (`Badge`) uses the `container` backend and it has a `validation` in place -- in the process of upgrading to rails 8 I had to upgrade mobility (1.2.9 > 1.3.x) and specs started failing so I'm going down the rabbit hole 

before opening an issue I wanted to come to grips with what might be happening and out of habit I decided to comment in place -- bad call maybe :)

doits: You are right this behaviour is wrong when cache plugin is used.

I guess I used `read` here just as a shortcut not to write it out what `#read` just above does. Which breaks when `#read` is overwritten by plugins, though!

And you are right `value` is never modified by this container backend. So it doesn't need to be reread after setting it and can be returned directly.

I added a test and implemented it in #671

---

Besides all this: Do you even need the `cache` plugin with the container backend? The container backend loads all data in a hash anyways when the model is loaded (or to be more precise active record loads it), so why the `cache` plugin?

dramalho: I really didn't need it, turns out we had it set as a default plugin in an initializer from way back when, I only noticed because specs started breaking when working on upgrading to rails 8 - which forced me to update mobility 

I did turn it off and, kinda spoiler alert, there is another issue cropping up with mobility, but I want to dig into it and find out what it is first 😅

ngobin: Thank you @dramalho! We just went through the Rails 8 and Mobility 1.2.9 -> 1.3.2 upgrade, and this issue broke a lot of tests. We use the jsonb container back end and disabling the `cache` plugin (which, like you, we had left on by default) immediately solves the problem. This thread made finding and diagnosing painless.

PR #567, opened 2022-04-14
instance variables will not get shared with subclasses, such as when …

### LINKED ISSUE #566: Does not work with subclassing in production only - dangerous bug
Okay so this bit me because the bug is only present when `config.cache_classes = true` in Rails.
I have a model that is using STI in the normal ActiveRecord way (type column). I add mobility/translates to the base class as normal.

Now the subclasses' translations will work fine in development. But if you turn on cache_classes (as is with production), reading any translated attribute will throw `KeyError: No backend for: myattr` `Caused by KeyError: key not found: :myattr`. The backend classes is an empty Hash on the class.

Would appreciate help or fix on this soon as I had to totally revert my migration from globalize to mobility due to this.

mrbrdo:
…using ActiveRecord STI. this will break the functionality in all subclasses, as their @mobility_backend_classes will be empty. fix by using class variables, which are shared between subclasses.

### DIFF
--- a/lib/mobility/plugins/backend.rb
+++ b/lib/mobility/plugins/backend.rb
@@ -161,14 +161,14 @@ Defines:
 
         def inherited(klass)
           parent_classes = mobility_backend_classes.freeze # ensure backend classes are not modified after being inherited
-          klass.class_eval { @mobility_backend_classes = parent_classes.dup }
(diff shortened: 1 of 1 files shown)


mrbrdo:
nevermind, this is not the solution unfortunately.

PR #568, opened 2022-04-14
Fix issues with subclassing, such as when using ActiveRecord STI.

### LINKED ISSUE #566: Does not work with subclassing in production only - dangerous bug

mrbrdo:
Fixes mobility to work if subclassing a subclass of a mobility-enabled class, and looks up mobility_backend_classes later in execution (upon being called) to be more flexible.

fixes #566

### DIFF
--- a/lib/mobility/plugins/backend.rb
+++ b/lib/mobility/plugins/backend.rb
@@ -159,16 +159,22 @@ Defines:
           raise KeyError, "No backend for: #{name}"
         end
 
-        def inherited(klass)
-          parent_classes = mobility_backend_classes.freeze # ensure backend classes are not modified after being inherited
-          klass.class_eval { @mobility_backend_classes = parent_classes.dup }
-          super
-        end
-
         protected
 
         def mobility_backend_classes
-          @mobility_backend_classes ||= {}
+          if @mobility_backend_classes
+            @mobility_backend_classes
+          else
+            @mobility_backend_classes = {}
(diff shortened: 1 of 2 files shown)


shioyama:
Thanks for the PR! I've been very busy recently and will be busy for the next few weeks but I'll have a look asap.

shioyama:
Can you write a failing test as well to demonstrate the issue?

mrbrdo:
@shioyama I added a test for this, which fails on master but passes with this fix.
Writing a test for the other issue I described in #566 is really tricky, but it did fix that issue for me as well.

mrbrdo:
@shioyama so since I added the failing spec as requested, and it is passing with this fix, can you merge this?

shioyama:
@mrbrdo Thanks very much for this, I see the issue now and this is a good fix. Sorry for the long delay in responding! I'll rebase on master and 1-2 and ship this.

PR #577, opened 2022-06-18
Do not query same attribute more than once

### LINKED ISSUE #564: Queries are broken if translates is called multiple times
The query scopes (Klass.i18n) stop working correctly if translates is called twice in a model class:
```ruby
class Model < ActiveRecord::Base
  translates :name 
  translates :name 
end
```
(not that it makes sense to do that, but it can happen due to class reloading etc)
The scope then generates queries like: `WHERE "table_translations_en"."name" = 'Name' AND "table_translations_en"."name" IS NULL`, so it never returns any records.

The issue is that `translates` does an `include Mobility.translations_class.new` every time, so it will add two `#<Translations (Mobility::Backends::ActiveRecord::Table) @names=name>` into Klass.ancestors. Then the query patches enumerate the modules with `scope.model.ancestors.grep(::Mobility::Translations)` and this will return 2 instances instead of 1, which results in adding bogus conditions to the query.

shioyama:
Fixes #564

### DIFF
--- a/spec/mobility/plugins/active_record/query_spec.rb
+++ b/spec/mobility/plugins/active_record/query_spec.rb
@@ -192,4 +192,29 @@ describe Mobility::Plugins::ActiveRecord::Query, orm: :active_record, type: :plu
       expect { query.order(:car_id) }.not_to raise_error
     end
   end
+
+  describe "regression for #564" do
+    it "works if translates is called multiple times" do
+      stub_const 'Article', Class.new(ActiveRecord::Base)
+      2.times { translates Article, :title, backend: :table }
+
+      article = Article.create(title: "Title")
+
+      expect(Article.i18n.where(title: "Title")).to eq([article])
+    end
+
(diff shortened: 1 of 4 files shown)


shioyama:
@mrbrdo This should fix #564, have a look. If it looks good I'll make a 1.2.7 patch and release that right away.

spec/mobility/plugins/active_record/query_spec.rb
```diff
@@ -192,4 +192,29 @@
...
+
+      expect(Article.i18n.where(title: "Title")).to eq([article])
+    end
+
+    it "handles intersecting attribute declarations" do
+      stub_const 'Article', Class.new(ActiveRecord::Base)
+      translates Article, :title, :content, backend: :key_value, type: :string
+
+      # title defined below clobbers title defined above
+      translates Article, :title, backend: :table
```
shioyama, line 211: Maybe we should issue a warning, but for now this at least handles the situation gracefully.

PR #582, opened 2022-06-22
Fix rails7 default scope issue with select

borisrorsvort:
We did not investigate too much but after upgrading to rails 7 we had issues when using `default_scope { I18n }` coupled with active record select method.

This PR seems to fix it but feel free to change and or add tests for other use cases. 

This seems to fix issues described in #513 back in the days

### DIFF
--- a/lib/mobility/plugins/active_record/query.rb
+++ b/lib/mobility/plugins/active_record/query.rb
@@ -137,6 +137,8 @@ enabled for any one attribute on the model.
           end
 
           def order(opts, *rest)
+            return super unless @klass.respond_to?(:mobility_attribute?)
+
             case opts
             when Symbol, String
               @klass.mobility_attribute?(opts) ? order({ opts => :asc }, *rest) : super
@@ -160,6 +162,8 @@ enabled for any one attribute on the model.
             %w[pluck group select].each do |method_name|
               define_method method_name do |*attrs, &block|
                 return super(*attrs, &block) if (method_name == 'select' && block.present?)
+
+                return super(*attrs, &block) unless @klass.respond_to?(:mobility_attribute?)
 
                 return super(*attrs, &block) unless attrs.any?(&@klass.method(:mobility_attribute?))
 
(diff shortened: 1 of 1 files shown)


borisrorsvort:
Currently this PR causes some issues in our rspec tests: 

```ruby
expected ActiveRecord::RecordNotFound, got #<NoMethodError: undefined method `[]=' for nil:NilClass> with backtrace:
           # /Users/ghost/Sites/mobility/lib/mobility/backends/active_record/pg_hash.rb:26:in `write'
           # /Users/ghost/Sites/mobility/lib/mobility/plugins/cache.rb:78:in `write'
           # /Users/ghost/Sites/mobility/lib/mobility/plugins/writer.rb:29:in `name='
           # /Users/ghost/Sites/mobility/lib/mobility/plugins/locale_accessors.rb:68:in `public_send'
           # /Users/ghost/Sites/mobility/lib/mobility/plugins/locale_accessors.rb:68:in `name_en='
```

shioyama:
I'd really like to reproduce the issue here before shipping a fix.

There's already a regression spec for #513 that runs on Rails 7:

https://github.com/shioyama/mobility/blob/b3420746e0ad7b2dea1d7fe13f2da6c513b93675/spec/mobility/plugins/active_record/query_spec.rb#L157-L194

It is passing so I'm not sure what's happening in your case. I don't doubt you're seeing the error but if the spec is not testing the problem correctly I'd like to fix that too.

shioyama:
Also:

> `default_scope { I18n }`

I assume you mean:

> `default_scope { i18n }`

right? (`i18n` not `I18n`)

shioyama:
> Currently this PR causes some issues in our rspec tests:

That doesn't look related to anything here.

shioyama:
A fix for this was released in [1.2](https://rubygems.org/gems/mobility/versions/1.2.9), closing this. Thanks for the issue report and fix!

borisrorsvort:
@shioyama awesome! ❤️

PR #588, opened 2022-08-12
Add column_fallback to Mobility initializer

fluke:
This feature was implemented in 1.2 but I had to read the PR to understand how to use it. Thought I'd add it into the default initializer.


shioyama:
Thank you! :heart:

PR #589, opened 2022-09-16
Make translatable references not nullable in rails migration generators

ralfclaassens:
Make translatable_id + translatable_type not nullable in rails migration generators


shioyama:
Thanks, this seems reasonable. I'm trying to think if there was any reason they weren't `null: false` originally... :thinking:

PR #590, opened 2022-09-19
add writer plugin type checking option

GeorgeGorbanev:
Hello! 

Some time ago I have discovered an issue with one of my job rails project. It was translated with Mobility using Postgres jsonb backend. So one of model attributes was used as one type in some places and different in another. And it was hard to figure out what type it was originally 😬 

I thought if with jsonb we losing traditional active record type castings and sql checkings why dont to add some basic checkings on library level? Of course it would be nice if my specific problem was prevented by properly testing or `.rbs` annotations or even just human comment after `translates` expression... but it was't. And I guess not only in my project.

So I decided to propose you type checking feature in writer plugin. I think this tiny but intelligent option will encourage developers to make them models more safely, explicit and reliable. Even when they can't for some reason write tests and types annotations :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -850,6 +850,61 @@ backend-independent queries on translated and untranslated attributes without
 having to deal with the details of how these translations are stored. The same
 interface is supported with Sequel to build datasets.
 
+### Optional type checking
+
+Since Mobility supports different storage backends some of them will not
+guarantee type safety of input data. For example Postgres
+([`Mobility::Backend::Jsonb`](http://www.rubydoc.info/gems/mobility/Mobility/Backends/Jsonb)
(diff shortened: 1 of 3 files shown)


GeorgeGorbanev:
@shioyama 👋

shioyama:
Sorry Mobility is basically feature-complete, I'm not adding any new functionality, only fixing actual bugs or incompatibility. You're welcome to implement this yourself by creating a plugin that inherits from `Mobility::Plugins::Writer`.

shioyama:
Thank you by the way, you made a very thorough PR :slightly_smiling_face: Unfortunately accepting changes also means maintaining them, and I have limited capacity to maintain what Mobility already has in it.

PR #604, opened 2023-02-20
Add attribute_names support to attribute_methods plugin

ippachi:
Thank you great gem ❤️ 

I want to use attribute_names.
And I think good location that this definition is located attribute_methods plugin.

What do you think?


ippachi:
Sorry I notice this changes make many cases break.
So I close this.

PR #615, opened 2023-05-27
Fix warning about freezing test double

shioyama:
This fixes a warning about freezing a test double.


shioyama:
Sequel builds are all failing, something must have changed.

Issue #544, opened 2021-11-09
Performance much slower in version 1.2.3


f1sherman:
Performance appears to be much slower in version 1.2.3 than version 1.2.1

## Context
I haven't been able to track this down completely but I believe the performance issue is related to [this PR](https://github.com/shioyama/mobility/pull/531) as I can't reproduce it on version 1.2.1 and we make extensive use of fallbacks.

I've tried to reproduce this with a vanilla Rails app and haven't been able to get but wanted to get this over to you in case you had any ideas.

Holding everything else constant, just initializing an ActiveRecord instance that has translated attributes is orders of magnitude slower.

Version 1.2.3:

```
Benchmark.realtime { 1000.times { Thing.new.title } }

=> 85.30171160000464
```

Version 1.2.1:

```
Benchmark.realtime { 1000.times { Thing.new.title } }

=> 0.3301175000015064
```

## Expected Behavior
Performance is not significantly worse in version 1.2.3 than 1.2.1

## Actual Behavior
Performance is significantly worse in version 1.2.3 than 1.2.1

## Possible Fix

Revert https://github.com/shioyama/mobility/pull/531

shioyama:
Hmm that's strange... the intention was to make it faster. I can revert but let me see if anything stands out. :eyes:

shioyama:
Ok I see the issue, I think #546 should fix it but I haven't benchmarked it. Could you have a look?

f1sherman:
@shioyama thank you! I tried that branch and I get an error when running my benchmark. Stack trace is below:

```
Exception: NoMethodError: undefined method `[]' for nil:NilClass
--
0: vendor/bundle/ruby/3.0.0/bundler/gems/mobility-15be6e8c2950/lib/mobility/backends/active_record/pg_hash.rb:19:in `read'
1: vendor/bundle/ruby/3.0.0/bundler/gems/mobility-15be6e8c2950/lib/mobility/plugins/active_record/query.rb:64:in `read'
2: vendor/bundle/ruby/3.0.0/bundler/gems/mobility-15be6e8c2950/lib/mobility/plugins/cache.rb:69:in `read'
3: vendor/bundle/ruby/3.0.0/bundler/gems/mobility-15be6e8c2950/lib/mobility/plugins/fallbacks.rb:156:in `block in read'
4: vendor/bundle/ruby/3.0.0/bundler/gems/mobility-15be6e8c2950/lib/mobility/plugins/fallbacks.rb:155:in `each'
```

[...]

PR #619, opened 2023-06-02
MySQL JSON backend

timkrins:
We have been using mobility for a few years using the JSON backend with a MySQL 8 JSON column.

I see a few people over the years have put together PRs, but they are still marked WIP.

The thing I am most unsure about in this PR is the `visit_Mobility_Plugins_Arel_Nodes_JsonDashDoubleArrow` function.
Maybe moving that quoting somewhere else would be better - I notice PR #271 from a few years ago has it elsewhere.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -40,13 +40,13 @@ namespace :db do
   desc "Set up the database schema"
   task up: :setup do
     orm = ENV['ORM']
-    return unless orm
-
-    require orm
-    require "database"
-    DB = Mobility::Test::Database.connect(orm)
-    require "#{orm}/schema"
-    Mobility::Test::Schema.up
+    if orm
+      require orm
+      require "database"
+      DB = Mobility::Test::Database.connect(orm)
+      require "#{orm}/schema"
+      Mobility::Test::Schema.up
+    end
   end
 
   desc "Drop and recreate the database schema"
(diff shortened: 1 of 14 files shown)


shioyama:
Thanks very much for sharing this! I noticed you pulled the commit fixing the Sequel failures, I need to fix that (it "fixes" the failures but doesn't really doing it correctly.)

Once I do that I'll have a look at this.

shioyama:
> but Rails 4.2.5.1 is 7 years old now...

Oh yeah I thought I'd removed that one from the builds. I'll remove it so don't worry about that :+1: Probably should also remove Rails 5 too actually.

timkrins:
I did pull that branch, as I noticed the tests were green, and wanted to run the full CI matrix on this PR.

I actually had a quick look at adding Sequel support for the JSON column too, and not just adding the ActiveRecord support, but I've never used Sequel and it doesn't seem to support JSON columns without another gem.

timkrins:
I just noticed some discussion on another issue about the default column value `{}` being *required* in JSON stores...
MySQL doesn't have defaults for JSON columns.
What should I be doing here? I haven't noticed any particular problem with not having a default but I am aware there might be something that I'm missing.
Possibly related to
https://github.com/shioyama/mobility/wiki/Postgres-Backends-%28Column-Attribute%29#railsactiverecord
(but it doesn't explain the reasons _why_ it is required)
and and issue mentioning this default
https://github.com/shioyama/mobility/issues/544

mpapis:
any updates on this?

PR #627, opened 2023-08-18
Don't try to load generators if Rails is loaded but ActiveRecord is not.

flop:
The generators are implicitly dependent on ActiveRecord, so if you are using Rails and you are not using ActiveRecord (We use Sequel with Rails), mobility would not load correctly and we would get the following error :

```
`require': cannot load such file -- rails/generators/active_record (LoadError)
```

This patch explicitly check that ActiveRecord is loaded before requiring the generators.


shioyama:
Can you rebase this on master? Should be green then.

shioyama:
I cherry-picked the first commit but not the dirty accessors one.

flop:
Thanks. Sorry for the second commit.. It was added automatically when I was trying to fix something else. I should have created a specific branch.

PR #634, opened 2024-01-31
Fix foreign_key containing the schema in the active record table backend

divanburger:
When a module is in a schema in the database the table name becomes something like 'my_schema.items' but the foreign key should not contain the schema part like it does now. So instead of the foreign key being 'my_schema.item_id' it will be 'item_id' instead.


shioyama:
Thanks! Can you add a test for this?

shioyama:
Test added in 6398256

PR #637, opened 2024-03-13
Fix passing unpacked options

t0biasz:
The define_handler_methods method in the Mobility gem's plugin was not properly handling method invocation with additional parameters, causing issues with methods like `will_save_change_to_attribute?(from:,to:)`. This fix modifies the else block to correctly pass additional parameters to the super method, ensuring proper functionality.

### DIFF
--- a/lib/mobility/plugins/active_model/dirty.rb
+++ b/lib/mobility/plugins/active_model/dirty.rb
@@ -142,7 +142,8 @@ the ActiveRecord dirty plugin for more information.
                     mutations_from_mobility.attribute_previously_changed?(attr_name))
                   mutations_from_mobility.send(#{method_name.inspect}, attr_name, *rest#{kwargs})
                 else
-                  super
+                  options = rest.first || {}
+                  super(attr_name, **options)
                 end
               end
               EOM


shioyama:
Can you rebase (I fixed some failing sequel tests) and add a failing test to show what you are fixing?

lib/mobility/plugins/active_model/dirty.rb
```diff
@@ -142,7 +142,8 @@ def #{method_name}(attr_name, *rest#{kwargs})
                     mutations_from_mobility.attribute_previously_changed?(attr_name))
                   mutations_from_mobility.send(#{method_name.inspect}, attr_name, *rest#{kwargs})
                 else
-                  super
+                  options = rest.first || {}
+                  super(attr_name, **options)
```
shioyama, line 146: Hmm, this doesn't look right. There is already some code handling keyword arguments above. Are you sure this fixes the issue? Please add a failing test.

doits:
I got something similar with `*_previously_changed?` methods and fixed it in #639. This feels similar to it. I can take a look.

doits:
I think I got it fixed correctly in #639.

shioyama:
Merged @doits's fix, thanks!

PR #643, opened 2024-07-23
Add support for to_json

bonekost:
This PR fixes a regression that I noticed in ActiveRecord::Base#as_json return value. (same as #630)

**Description**

If you enable attribute_methods plugin, mobility patches ActiveRecord::Base#attributes method to include translated attributes as wekk. The same change used to apply to ActiveRecord::Base#as_json too.

But the internal implementation of serialization [changed slightly in Rails 7.0](https://github.com/rails/rails/commit/322e62842ce656e4bb7b6562efc2a7f314fe5d43#diff-89d44807164a1461ee3c9a225c9f1301a87fdac8b45b4759b035d37157eced14R25) to improve the performance.

As a result, as_json method no longer includes translated attributes.

**Solution**

I added an override of attribute_names_for_serialization method into the plugin to make it include the missing translated fields into serialized_attributes hash which is used by as_json later.

attribute_names_for_serialization would be never be called by Rails 6.x or lower, but I still added a safe guard against calling super.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,13 @@
 
 ## 1.3
 
-### Unreleased
+### 1.3.1
 
 - Use classify to generate table name
   ([#634](https://github.com/shioyama/mobility/pull/634)), thanks
   [divanburger](https://github.com/divanburger)!
+- Fix select + count in ActiveRecord
+  8 ([#659](https://github.com/shioyama/mobility/pull/659))
 
 ### 1.3.0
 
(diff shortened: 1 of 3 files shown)


ollym:
@shioyama this is a rebase + changelog for https://github.com/shioyama/mobility/pull/630 as that PR seemed to stop

shioyama:
Sorry haven't had much time to look at this repo for a while, thanks for the fix! Can you rebase?

bonekost:
@shioyama I rebased and re-opened new PR #660 for this json support

PR #646, opened 2024-08-21
Fix fallback syntax in initializer example

pwim:
The previous sample code passed a block to the fallbacks method. Instead, they should be an argument.


shioyama:
Thanks!

Issue #652, opened 2024-11-09
Add support for Rails 8


n-rodriguez:
Hi there!

I've updated my app to Rails 8 and now got this error in CI :

```sh
     NoMethodError:
       undefined method `mobility_backend_class' for nil
     # /cache/nicolas/concerto/ruby-wip-rails8/ruby/3.3.0/gems/mobility-1.2.9/lib/mobility/plugins/active_record/query.rb:192:in `backend_node'
     # /cache/nicolas/concerto/ruby-wip-rails8/ruby/3.3.0/gems/mobility-1.2.9/lib/mobility/plugins/active_record/query.rb:233:in `block (2 levels) in _build'
     # /cache/nicolas/concerto/ruby-wip-rails8/ruby/3.3.0/gems/mobility-1.2.9/lib/mobility/plugins/active_record/query.rb:232:in `map'
     # /cache/nicolas/concerto/ruby-wip-rails8/ruby/3.3.0/gems/mobility-1.2.9/lib/mobility/plugins/active_record/query.rb:232:in `block in _build'
     # /cache/nicolas/concerto/ruby-wip-rails8/ruby/3.3.0/gems/mobility-1.2.9/lib/mobility/plugins/active_record/query.rb:227:in `each'
     # /cache/nicolas/concerto/ruby-wip-rails8/ruby/3.3.0/gems/mobility-1.2.9/lib/mobility/plugins/active_record/query.rb:227:in `inject'
     # /cache/nicolas/concerto/ruby-wip-rails8/ruby/3.3.0/gems/mobility-1.2.9/lib/mobility/plugins/active_record/query.rb:227:in `_build'
     # /cache/nicolas/concerto/ruby-wip-rails8/ruby/3.3.0/gems/mobility-1.2.9/lib/mobility/plugins/active_record/query.rb:213:in `build'
     # /cache/nicolas/concerto/ruby-wip-rails8/ruby/3.3.0/gems/mobility-1.2.9/lib/mobility/plugins/active_record/query.rb:130:in `where!'
     # /cache/nicolas/concerto/ruby-wip-rails8/ruby/3.3.0/gems/activerecord-8.0.0/lib/active_record/relation/query_methods.rb:1039:in `where'
     # /cache/nicolas/concerto/ruby-wip-rails8/ruby/3.3.0/gems/mobility-1.2.9/lib/mobility/plugins/active_record/query.rb:136:in `where'
[...]

PR #654, opened 2024-11-26
Fixes for ActiveRecord 8.0

shioyama:
https://github.com/shioyama/mobility/issues/652

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ group :development, :test do
   when 'active_record'
     orm_version ||= '7.0'
     case orm_version
-    when '6.1', '7.0', '7.1'
+    when '6.1', '7.0', '7.1', '8.0'
       gem 'activerecord', "~> #{orm_version}.0"
     when 'edge'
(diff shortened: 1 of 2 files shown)


jukra:
@shioyama I think this caused the second issue with Rails 8 and the AS statement not getting wrapped correctly:
https://github.com/rails/rails/commit/ba468db0bdc880c694df091b5800d114e963eff0

Previously we got the `Arel::Nodes` for `column_name` in `execute_simple_calculation`. With Rails 8 we actually get just the string `"article_translations_en"."title" AS __mobility_title_en__,` (https://github.com/rails/rails/blob/8-0-stable/activerecord/lib/active_record/relation/calculations.rb#L445)

This is why the `count` in `Countable` module does not get called at all (https://github.com/shioyama/mobility/blob/master/lib/mobility/plugins/arel.rb#L32)

If you monkey patch the `select_for_count` from Rails 7.2 to 8.0, it works, but I don't think that's the correct way to fix it.

PR #658, opened 2024-11-30
Use explicit default for configuration

shioyama:
A few years later, I find the "magical" implict receiver format used in the curent readme kind of weird:

```ruby
Mobility.configure do
  backend :key_value

  ...
end
```

Mobility actually also supports a more explicit variant:

```ruby
Mobility.configure do |config|
  config.plugin(:backend, :key_value)

  ...
end
```

I kind of prefer the latter. Although it's more characters to type (and a lot of "plugin") I think it's actually more readable.

What do folks think?


shioyama:
I also really don't like this kind of thing:

```ruby
  config.plugin(:dirty, false)
```

What the `false` means here is super obscure. But I don't know how to improve this without breaking backwards compatibility :thinking:

Issue #664, opened 2024-12-19
Cache returns last value if nil value provided


unavailabl3:
## Context
When assigning attribute (which uses translations) `nil` value, it won't invalidate cache and will store previous value.
This possibly can affect any logic, but the common example is `ActiveRecord` validations. As cache uses previous value, it will pass `presence` validation, but it is incorrect as current value is nil.
Disabling cache for the attribute fixes situation, but actually it is not a solution.
This change appeared after version 1.2.9 and possibly after implementing logic for the Issue #535  

- Mobility 1.3.1
- Ruby 3.3.5
- Rails 7.2
- ActiveRecord
- Postgresql 15.7
- Minitest
- FactoryBot

```
class Item < ApplicationRecord
  extend Mobility
  translates :name

  validates :name, presence: true
end
```

## Expected Behavior
Cache should be invalidated if new value is `nil` for translatable attribute
```
test 'should not save when name empty' do
  item = FactoryBot.build(:item)
  item.name # 'some-fake-text'
  item.name = nil
  item.name # nil
  item.inspect # ...name: {},...
  assert_not item.save
end
```

## Actual Behavior
```
ActiveRecord::PgHash -->
...
def write(locale, value, _options = nil)
  if value.nil?
    translations.delete(locale.to_s) # returns previous value and will be used by Cache->write method
  else
    translations[locale.to_s] = value
  end
end
```
Attribute shows the previous cached value because of `.delete()` method which returns removed value
```
test 'should not save when name empty' do
  item = FactoryBot.build(:item)
  item.name # 'some-fake-text'
  item.name = nil
  item.name # 'some-fake-text' <-- wrong, previous value from cache
  item.inspect # ...name: {},...  <-- hash update is correct
  assert_not item.save <-- fail, passes name validation because of cached value
end
```

## Possible Fix
return `nil` value if translation was deleted from hash in `mobility/backends/active_record/pg_hash.rb`
```
def write(locale, value, _options = nil)
  if value.nil?
    translations.delete(locale.to_s)
    nil # <-- return nil which will be used by Cache -> write method
  else
    translations[locale.to_s] = value
  end
end
```
This short fix works for my projects, but it is better to verify other ORMs or underlying logic of the gem.
[Opened Pull Request](https://github.com/shioyama/mobility/pull/665)

shioyama:
Thanks for the detailed write-up! I merged #665. I'll leave this open for now so I can have a closer look before closing.

shioyama:
[...]

PR #665, opened 2024-12-19
Issue #664 | Update handling nil value for PgHash

unavailabl3:
## [Issue 664: Cache returns last value if nil value provided](https://github.com/shioyama/mobility/issues/664)

When assigning attribute (which uses translations) nil value, it won't invalidate cache and will store previous value.
This possibly can affect any logic, but the common example is ActiveRecord validations. As cache uses previous value, it will pass presence validation, but it is incorrect as current value is nil.
Disabling cache for the attribute fixes situation, but actually it is not a solution.
This change appeared after version 1.2.9 and possibly after implementing logic for the Issue https://github.com/shioyama/mobility/issues/535


shioyama:
Thanks! :heart:

PR #671, opened 2025-02-09
Simply return `value` after write in active_record container backend

doits:
Simply return `value` after write in active_record container backend

This fixes a bug where `#read` is overwritten by the cache plugin and
therefore the container backend returns stale data. It should always
return the value it wrote.

Since `value` is never modified by this backend it makes no sense to
read it again after setting it.

Besides fixing the bug it might give a small perfomance boost, too.

fixes https://github.com/shioyama/mobility/pull/543/files#r1944637671

thanks @dramalho for uncovering the bug and hints

### DIFF
--- a/lib/mobility/backends/active_record/container.rb
+++ b/lib/mobility/backends/active_record/container.rb
@@ -37,7 +37,7 @@ Implements the {Mobility::Backends::Container} backend for ActiveRecord models.
       # @return [String,Integer,Boolean] Updated value
       def write(locale, value, _ = nil)
         set_attribute_translation(locale, value)
-        read(locale)
+        value
       end
       # @!endgroup
 
--- a/spec/mobility/backends/active_record/container_spec.rb
+++ b/spec/mobility/backends/active_record/container_spec.rb
@@ -52,6 +52,17 @@ describe "Mobility::Backends::ActiveRecord::Container", orm: :active_record, db:
     end
   end
 
+  context "with cache plugin" do
+    plugins :active_record, :reader, :writer, :cache
+    before { translates ContainerPost, :title, backend: :container }
+
+    it 'resets cache on write' do
+      post = ContainerPost.create!
+
+      expect { post.title = 'aaa' }.to change { post.title }.from(nil).to('aaa')
+    end
+  end
+
   context "with query plugin" do
     plugins :active_record, :reader, :writer, :query
     before { translates ContainerPost, :title, :content, backend: :container }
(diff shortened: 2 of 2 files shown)


dramalho:
Lovely, thank you ☺️

lib/mobility/backends/active_record/container.rb
```diff
@@ -37,7 +37,7 @@ def read(locale, _ = nil)
       # @return [String,Integer,Boolean] Updated value
       def write(locale, value, _ = nil)
         set_attribute_translation(locale, value)
-        read(locale)
+        value
```
shioyama, line 40: This will sometimes return the wrong value, e.g. if you have the presence plugin enabled and assign `""`, this would return `""` whereas it should return `nil`.

spec/mobility/backends/active_record/container_spec.rb
```diff
@@ -52,6 +52,17 @@
...
+  context "with cache plugin" do
+    plugins :active_record, :reader, :writer, :cache
+    before { translates ContainerPost, :title, backend: :container }
+
+    it 'resets cache on write' do
+      post = ContainerPost.create!
+
+      expect { post.title = 'aaa' }.to change { post.title }.from(nil).to('aaa')
+    end
+  end
```
shioyama, line 64: Thank you for including a test!

shioyama:
This seems reasonable to me, my only concern with returning `value` directly is how this would interact with e.g. the presence plugin. Currently I believe if you have the plugin enabled and assign `""`, you will get back `nil`, whereas with this change I think you will get back `""`. Can you check?

doits:
> Currently I believe if you have the plugin enabled and assign "", you will get back nil, whereas with this change I think you will get back ""

I don't think this is how ruby works with setters if I am correct, e.g. any setter always returns the passed in value no matter what:

```ruby
module Something
  def self.bla
    puts "calling getter"

    @bla
  end

  def self.bla=(value)
    puts "calling setter"

    @bla = value

    # try to return something different from setter
    "something different"
  end
end

Something.bla
# calling getter
# => nil

Something.bla = 'my value'
# calling setter
# => "my value"

Something.bla
# calling getter
# => "my value"
```

Or did you think of something different?

---

I added a spec though to check that the presence plugin is applied to the value written to the backend, e.g. it does not save `""` to the the database. This still works because `#write` is overwritten by the plugin so the call chain still ensures the the presence plugin does its work (`ActiveRecord::Container#write` already gets the `value` which was modified by the presence plugin).

AndyStabler:
Thanks for this fix @doits 🙇🏻 It's always nice when someone has already fixed what seemed to be a very specific issue. 😄 Any chance we could get this merged in @shioyama ?

PR #677, opened 2025-08-28
use translation model's connection pool

ollym:
For situations where you might want to leverage Rails' multiple databases, eg.

```ruby
class ApplicationRecord < ActiveRecord::Base
  self.abstract_class = true

  connects_to database: { writing: :primary, reading: :primary_replica }
end
```

```ruby
class Word < ApplicationRecord
  extend Mobility
  translates :name, :meaning
end
```

Currently if you do this:

```ruby
ApplicationRecord.connected_to(role: :reading) do
  Word.first.name
end
```

The word record is fetched using the reading role, but the translation is fetched using the writing role because the Translation class inherits from it's own abstract class.

This PR introduces some logic that will use by default the abstract class of the model that it is translating, and therefore reuse its connection properties.

### DIFF
--- a/lib/mobility/backends/active_record/table.rb
+++ b/lib/mobility/backends/active_record/table.rb
@@ -250,7 +250,11 @@ columns to that table.
           if self.const_defined?(subclass_name, false)
             const_get(subclass_name, false)
           else
-            const_set(subclass_name, Class.new(Translation))
+            const_set(subclass_name, Class.new(Translation) {
+              def self.connection_pool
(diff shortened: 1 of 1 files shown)


ollym:
@shioyama any chance of getting this merged? This fixes how it works with Rails' multiple databases. Rails internally adopts the same mechanism for habtm tables here: https://github.com/rails/rails/blob/main/activerecord/lib/active_record/associations/builder/has_and_belongs_to_many.rb#L45-L47

Thanks

